PR #6, opened 2014-01-18
Update build.xml to support ld --as-needed

Ampelbein:
Ubuntu uses the linker option "--as-needed" by default. This requires that libraries are added in the correct order to the linker command line (i.e. after the objects requesting external symbols). This change fixes the build on Ubuntu.


ibauersachs:
This was independently added in commit 0d0cd062333bc76525732a2f088d9c166ae05e1f.

PR #19, opened 2014-09-18
Created pom.xml

mondain:
The downside is that some of the libraries must be provided in project lib directory because they are unavailable from a repo and / or they are specially modified for jitsi.


stokito:
I started working on mavenization. I decided to split java sources to one artifact and create another artifacts for each native lib that should be compiled to so and dll.

Here is java part https://github.com/stokito/libjitsi/tree/maven
Native part https://github.com/stokito/libjitsi/tree/maven-native

Then I found that some libraries may be duplicated with other JNI bindings.
For example, official VPX repository already have JNI binding, but it looks obsolete https://code.google.com/p/webm/source/browse/?repo=bindings
So I decided to make it as separate repostitory [libjnvpx](https://github.com/stokito/libjnvpx) and merge with official binding.

bgrozev:
Maven is now supported with @lyubomir 's 8c139bf0f1784eca2da01bf48b8349ce60d99135.

PR #24, opened 2014-12-04
Added certificate signature algorithm override for DTLS

mondain:
Added support for overriding the x509 certificate signature algorithm via system property when use in DTLS with WebRTC. If the system property is not overridden, the "old" value of SHA1withRSA is used. Allowing this to be overridden provides better support for Chrome and Firefox. To use the override, add this Java option to your startup: -Dorg.jitsi.impl.transform.dtls.SIGNATURE_ALGORITHM=SHA256withRSA


mondain:
Sunsetting Sha-1 http://googleonlinesecurity.blogspot.ca/2014/09/gradually-sunsetting-sha-1.html

ibauersachs:
I still don't mind having this configurable, but not from a system property. Please use the config service as mentioned before. And while you're at it, please respect the coding guidelines with the line length limit of 80 chars.

mondain:
I don't get why you all format the way that you do; are you guys using old 40 column CRT monitors? Irregardless, I have been trying to format your way with these pulls, so it must have slipped through the cracks. Lastly, I've done this pull twice already, so if the config service is the final way to go I'll be glad to modify this a third time. Keep up the excellent work.

ibauersachs:
If you're still interested in bringing this up-to-date, please do.
- I don't know how 80 chars came into existence originally, AFAIK its a preference of Emil. For new code the point is consistency. Personally I don't like it either.
- Sorry for doing so much work about this simple change. But yes, the config service is the preferred way to override the default if the service is available.

mondain:
I'll try to get an update for this in the next few days.

ibauersachs:
Thanks!
There seem to be conflicts now, can you please rebase your patch on master? Also there are still some formatting issues and the config service needs to be null-checked before accessed.

mondain:
Ok, will do.

mondain:
I used rebase, if this seems confusing @ibauersachs, then I will gladly resubmit this from a new "cleaner" fork.

ibauersachs:
This seems like a merge rather a rebase - and it still conflicts. Please also squash your commits. Otherwise the actual change looks okay now.

mondain:
I'm gonna close and recreate this from master, my commit is fairly simple and doesn't need all this extra baggage.

PR #27, opened 2014-12-10
Modified audio channel retrieval

mondain:
Forcing channels to Format.NOT_SPECIFIED negates the check in the "if" block on line 1594. Getting the specified channels from the AudioFormat would seem to be more effective.


ibauersachs:
@lyubomir can you please take a look at this? It seems sensible to me, but I don't know this code and the effects of it.

PR #64, opened 2015-10-12
Adds a DebugTransformEngine.

gpolitis:
I had this lying around for a while and I thought that it might be helpful for somebody else. It requires some change in the PacketLoggingServiceImplementation as well so that it actually logs ARBITRARY packets.

You can setup logging like this:
```
net.java.sip.communicator.packetlogging.PACKET_LOGGING_ENABLED=true
net.java.sip.communicator.packetlogging.PACKET_LOGGING_ARBITRARY_ENABLED=true
net.java.sip.communicator.packetlogging.PACKET_LOGGING_FILE_COUNT=1
net.java.sip.communicator.packetlogging.PACKET_LOGGING_FILE_SIZE=-1
```

You setup a named pipe like this:
```
mkfifo ~/.sip-communicator/log/jitsi0.pcap
```

You launch Wireshark like this:
```
wireshark -k -i <(cat ~/.sip-communicator/log/jitsi0.pcap)
```

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -300,8 +300,7 @@ public class MediaStreamImpl
      * The <tt>TransformEngine</tt> instance that logs packets going in and out
      * of this <tt>MediaStream</tt>.
      */
-    private final DebugTransformEngine debugTransformEngine
-        = new DebugTransformEngine(this);
+    private DebugTransformEngine debugTransformEngine;
 
     /**
      * The <tt>TransformEngine</tt> instance registered in the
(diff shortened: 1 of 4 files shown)


ibauersachs:
This does look useful.
Given how often this code is executed, I think it would make sense to move the enabled check to where the transformer is added to chain. To keep the property "confined" withn the DebugTransformer, you could use a static factory method that only returns an instance if enabled or simply expose the enabled through a property.

Please remove commented debug code and what is that // region stuff about?

gpolitis:
Thanks very much for your valuable input @ibauersachs. I'll update the PR with an improved engine based on your comments. About the "//region .." comments, it's an IntelliJ IDEA feature that allows the [folding of custom regions with line comments](https://www.jetbrains.com/idea/help/folding-custom-regions-with-line-comments.html). It's can be handy if you're using that IDE, but I can remove it if you have a strong preference against it.

gpolitis:
This PR is complemented by https://github.com/jitsi/jitsi/pull/171

ibauersachs:
C# has them as a language feature and there have been endless debates of whether to use them or not. Being an Eclipse user I guess you can guess my preference. Please check with your teammates at Atlassian.

gpolitis:
Thank you very much for sharing your thoughts on this Ingo. I have been using the controversial regions back when I was coding in C#, but I agree with you that they need to be removed from our code as it is non-standard practice.

PR #65, opened 2015-10-27
Added certification creation algorithm override via the Configuration…

mondain:
Added the means for overriding the certificate creation algorithm for DTLS using the ConfigurationService. 
This was PR #24 but has been recreated based on current master.


mondain:
Thanks for merging this, it was sorely needed in my current project.

PR #66, opened 2015-11-15
Use 2048 bit RSA keys for DTLS

bickelj:


mondain:
I'd suggest that this be done via the config service, like the algorithm selection.

bickelj:
Good suggestion, agree. I'm looking at your previous contribution as a model for doing it.

This PR could then be considered "make 2048 the default."

PR #68, opened 2015-12-02
Lowers the level of the multi-line message with stream statistics pri…

bgrozev:
…nted on MediaStream#close().


gpolitis:
Looks safe to me.

PR #72, opened 2016-01-08
Update boncycastle from 1.51 to 1.54

champtar:
This is the same version as in jitsi-universe, just setting it in the stone as I want to update to 1.54

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,10 +37,12 @@
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcpkix-jdk15on</artifactId>
+      <version>1.54</version>
     </dependency>
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcprov-jdk15on</artifactId>
+      <version>1.54</version>
     </dependency>
     <!-- org.jitsi -->
     <dependency>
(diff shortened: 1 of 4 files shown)


mondain:
I'm using BC 1.53 and it seems to work just fine; my 2 cents.

gpolitis:
We're actually building bouncycastle from this branch https://github.com/gpolitis/bc-java because it contains a fix required for Firefox support. So, in order to switch to 1.53 we need to merge the changes from upstream into the aforementioned branch, build the library and push to our GH maven repository.

gpolitis:
Actually @damencho just brought to my attention that DTLS version negotiation is supposed to be fixed in BouncyCastle 1.4. We need to test this new version against the bridge, and eventually switch to that if no other problem arises.

champtar:
I ve a patch to fix the build failure, i ll sens et tomorrow (it's on my work PC)

champtar:
Here is my patch. I'm using bouncy castle 1.54 in all jitsi's projects and i can join a room with firefox nightly (is it a good enough test?)

bickelj:
Just tested bc 1.54 in:
libjitsi
jitsi-videobridge (using above libjitsi)
jicofo (using above libjitsi)

These three on the jitsi-meet server, Abrowser 43.0.4 (Firefox) Trisquel GNU/Linux had successful video chat with Firefox 43.0.4 on another operating system.

Tests with jitsi client to jitsi client over videobridge were unsuccessful but that may have been due to other factors.

ibauersachs:
@gpolitis Any reason not to merge this?

gpolitis:
If we merge this, libjitsi will depend on 1.54 and jitsi-videobridge and jicofo on 1.51-jitsi-1. Since 1.54 has fixed DTLS negotiation, I don't see a reason why not to use 1.54 everywhere but maybe I'm missing something?

champtar:
Maybe we should update bc in jitsi-universe

PR #76, opened 2016-01-25
Sets a capacity of 100 to the LinkedBlockingQueue-s used to hold obje…

bgrozev:
…cts to be reused.

### DIFF
--- a/src/org/jitsi/impl/neomedia/RTPConnectorInputStream.java
+++ b/src/org/jitsi/impl/neomedia/RTPConnectorInputStream.java
@@ -175,7 +175,8 @@ public abstract class RTPConnectorInputStream<T>
      * The pool of <tt>RawPacket</tt> instances to reduce their allocations and
      * garbage collection.
      */
-    private final Queue<RawPacket> rawPacketPool = new LinkedBlockingQueue<>();
+    private final Queue<RawPacket> rawPacketPool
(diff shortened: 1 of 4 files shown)


mondain:
Might I suggest that you make the value a parameter using the Config service?

bgrozev:
Is this a general suggestion, or you see a case where you would need to tune it?

mondain:
I'd like the option to tune something that is set to create ~400 extra objects * number of streams I've got going.

bgrozev:
It will not create 400 objects for each stream, it will limit the number of these objects to 100 per queue. Until now there was practically no limit.

mondain:
It's true that what you've done is probably better than leaving it open-ended and possibly leading to an OOM error and while it doesn't create ~400 objects, it creates slots for them to be inserted per stream. If allowing the implementer some means to configure this is not applicable, at least you've shown me a spot extend these classes.

bgrozev:
I still don't see how this would be a useful control to have, and I don't have time to spend on it. If you'd like to implement it I would be happy to review and merge, though.

ibauersachs:
@bgrozev Can you please rebase this PR against current master? 900769c5a6967676e0054401a6b6dc2f5a47b448 introduced incompatible changes.

bgrozev:
Rebased. Also included Paul's suggestion.

PR #77, opened 2016-01-28
Use fixed Random seed values and a latch for test consistency.

bickelj:
Use independent Random instances with fixed seed values for consistency.
Also use the wait form described in JCIP 14.2.2.

Tried a couple of seed values that would still exercise the tests but complete in under a minute on my machine. The modification to the synchronizer, e.g. using wait in a loop with condition predicate being checked as described by Goetz et al in JCIP 14.2.2, might have fixed a bug.

Test consistently passes and in a reasonable amount of time.

### DIFF
--- a/test/org/jitsi/sctp4j/TestLink.java
+++ b/test/org/jitsi/sctp4j/TestLink.java
@@ -16,6 +16,7 @@
 package org.jitsi.sctp4j;
 
 import java.io.*;
+import java.util.*;
 
 /**
  * The link that loses some packets and produces IOExceptions from time to time.
@@ -28,6 +29,9 @@ public class TestLink
     private final double lossRate;
 
     private final double errorRate;
+
+    /** Set random seed for consistent tests */
+    private static final Random rand = new Random(1234567);
 
     public TestLink(SctpSocket a, SctpSocket b,
                     double lossRate, double errorRate)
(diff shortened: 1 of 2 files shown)


bickelj:
Can we try the changes in [500c6e9](https://github.com/bickelj/libjitsi/commit/500c6e94c1dbee44ee0784eeb36597d678d5aed5)? The timeout on my machine needed to be increased to above 5 seconds for test passage.

Also let me know what you think about adding a message to the user indicating that the IOExceptions are expected during the test run. The first time I ran these tests I was not sure if those were expected or not.

test/org/jitsi/sctp4j/TestLink.java
```diff
@@ -43,7 +47,7 @@ public TestLink(SctpSocket a, SctpSocket b,
     public void onConnOut(SctpSocket s, byte[] packet)
         throws IOException
     {
-        double r = Math.random();
+        double r = getRandomDouble();
```
ibauersachs: Call rand.nextDouble() directly here.

ibauersachs:
Can you please squash your commits and fix the [formatting](https://jitsi.org/Documentation/CodeConvention)? When done and unless @paweldomas has further objections I'm ready to merge this.

bickelj:
Did I just ruin this request's comment history? In future should I use a new branch instead of force pushing the existing branch?

ibauersachs:
Seems like it, even though my lone getDouble() is still there :(
http://forum.imagej.net/t/force-push-a-branch-without-losing-review-comments/1009/2

bickelj:
I force pushed the original branch (to see what would happen) and now the original comments are back. Magical.

Created https://github.com/jitsi/libjitsi/pull/96 as the successor to this PR.

PR #79, opened 2016-02-09
Fix OpenSSL JNI with recent libcrypto

champtar:


ibauersachs:
This is obsoleted by #83, right?

champtar:
not the first commit, i will rework it, please leave this PR open
also i haven't benchmarked the hmac part yet, maybe it's faster (or close enough) in Java

PR #81, opened 2016-02-15
Fix NullPointerException in case of rewriting simulcast mode without …

lluczo:
…red and fec


gpolitis:
Thank you for the fix! Before merging, we need you to sign our contributor agreement ([individual](https://jitsi.org/icla) or [corporate](https://jitsi.org/ccla)).

gpolitis:
Thank you @lluczo !

PR #83, opened 2016-02-17
Rework encryption openssl offloading

champtar:
This is a WIP, but it's working :)
-  drop F8 mode as it's optional and apparently designed for UMTS
- cleanup & rework SRTPCipherCTR (our current java AES CTR implementation)
- rename SRTPCipherCTR to SRTPCipherCTRJava, make SRTPCipherCTR an interface
- add SRTPCipherCTROpenSSL (1 JNI call to encrypt a packet instead of 1 for every 16 bytes)
- use dlopen to load libcrypto, this allow us to work with all libcrypto 1.0.x version (we only depend on the functions that we use) for now this is linux only but making it work on windows and mac shouldn't be hard
- remove OpenSSLBlockCipher

according to my benchmark SRTPCipherCTROpenSSL is 6 times faster than SRTPCipherCTRJava (with JCE provider), and now aes enc/dec is outside of the top 10 in perf-java-top

Comments welcome


ibauersachs:
Please remove the normalize commit, it shouldn't be necessary anymore.

I'm very reluctant to remove the F8 mode. It is a mandatory option for the SDES key exchange and AFAIK also for ZRTP. Remember that libjitsi isn't just for the Videobridge.

ibauersachs:
Also, please make sure that the code formatting is correct in all files. There are a lot of incorrect brace positions and lines that are too long.

Before you spend more work though, wait for confirmation from @lyubomir that replacing the BlockCipher with SrtpCipherCtrJava/OpenSsl is fine. It does make a lot of sense to me.

champtar:
@ibauersachs in RFC3711 AES-f8 is optionnal (https://tools.ietf.org/html/rfc3711#section-5), can you show me where it's mandatory?
I see no mention of AES-f8 in http://www.rfc-editor.org/rfc/rfc6189.txt (ZRTP)
It's in https://tools.ietf.org/html/rfc4568#section-6.2, but it's not mandatory, as AES GCM is not mandatory but do exists

Also i already asked on the ML (http://lists.jitsi.org/pipermail/dev/2016-February/026832.html) but nobody gave me a user of AES-f8, that's why i dropped it (and it simplify my rework)

ibauersachs:
RFC4568, section 6.2.3.

ibauersachs:
RFC4568 section 6.2 unfortunately lacks any definition of which cipher suites are mandatory. However I'd consider an implementation incomplete because it is defined in the original spec and not in an extension (like 256bit encryption or as you mentioned the gcm suites).

Lets discuss that in TheCall in an hour.

lyubomir:
@champtar, could you please explain why you closed this PR? I started reviewing it yesterday.

champtar:
I've a better version comming, where i don't drop F8, and do some more cleanup.
Doing final tests right now.
Many parts are similar so you haven't lost too much time,
sorry for the lack of communication.

PR #84, opened 2016-02-17
Increase the send-side bandwidth estimator max bitrate limit to 20Mbps.

bgrozev:


lyubomir:
Thank you, @bgrozev!

If you allow me, a couple of comments (which I think are out of the scope of this pull request):

- I didn't really understand why you increased the maxBitrate value. I did try searching the WebRTC source code in Chromium and I believe I found the old value but not the new value. It may make sense to add a javadoc stating why you raised this limit.

- The fields minBitrate and maxBitrate are static finals. I didn't find them in the WebRTC source code in Chromium so it doesn't look to me like you're trying to preserve outside names. Consequently, wouldn't it be consistent to have them in uppercase like it's usual for Java?

bgrozev:
Thanks, @lyubomir. The constants were originally based on constants from webrtc.org, but since we are changing them anyway, I renamed them in #85. The reason we increase the max is because we want our estimation to grow sufficiently in a scenario where we are sending (or forwarding) multiple streams.

PR #88, opened 2016-02-24
Add .gitattributes to prevent future EOL problems

champtar:
my script was only converting mixed EOL (CRLF and LF),
6 files remains with CRLF only, please run
find -name '*.java' | xargs file | grep 'CRLF' | cut -d ':' -f1 | xargs -n1 dos2unix

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.java text eol=lf diff=java


ibauersachs:
I'm always a bit confused how .gitattributes actually works. Shouldn't this be `*.java eol=lf`?

champtar:
no, it will store LF inside git, but let you use CRLF on windows and LF on linux

champtar:
https://help.github.com/articles/dealing-with-line-endings/

ibauersachs:
According to this link, it should be `*.java text eol=lf`. There is no reason to checkout files as CRLF, even on Windows.

champtar:
i'm not a windows user so i have no preferences here, we can force LF for *.java

champtar:
we can also add 'diff=java'

ibauersachs:
Not sure who else uses Windows, but I don't see the point of converting files to CRLF on checkout. It just makes it more likely to cause problems and except for Notepad all editors support LF anyway.

Adding diff is fine for me.

champtar:
i've updated my PRs

ibauersachs:
This PR doesn't have the changes. Did you forget you push?

champtar:
Hi @ibauersachs,
It's impossible to review such changes, so I let you do the dos2unix

ibauersachs:
Yes, I was referring to the content of the .gitattributes file of libjitsi (the one for Jitsi is merged and I'll look at the mixed content there now).

champtar:
Sorry, i must have forget to push (tomorrow)

champtar:
hi @ibauersachs,
done

champtar:
don't forget to cleanup
find -name '*.java' | xargs file | grep 'CRLF' | cut -d ':' -f1 | xargs -n1 dos2unix

PR #89, opened 2016-02-24
Uses the real RTT for send-side bandwidth estimation.

bgrozev:


src/org/jitsi/impl/neomedia/rtp/sendsidebandwidthestimation/SendSideBandwidthEstimation.java
```diff
@@ -366,10 +366,10 @@ public void rembReceived(long bitrateBps)
      * Returns the last calculated RTT to the endpoint.
      * @return the last calculated RTT to the endpoint.
      */
-    private synchronized int getRtt()
+    private synchronized long getRtt()
     {
-        //FIXME: the RTT from MediaStreamStats is wrong (always 2^16).
-        return 100;
+        long rtt = mediaStream.getMediaStreamStats().getRttMs();
```
gpolitis: Can we add some logging here to catch suspicious RTTs  : `rtt =< 0 || rtt >= 1000`?

bgrozev: Yes, added. Thanks!

PR #91, opened 2016-02-28
Rewrite simulcast timestamps via wallclocks

lyubomir:
Upon rewriting RTP packets for the purposes of SSRC rewriting, rewrites their RTP timestamps as well in order to produce an output RTP stream with monotonically increasing timestamps with a single timestamp base. The rewriting utilizes information about the remote wallcocks of the respective SSRCs (i.e. RTP streams).


gpolitis:
I'm testing this and in my (brilliant) setup (that reorders UDP packets..) a lot of packets are being dropped at the receiver as replay attacks.

gpolitis:
Note that I have no idea where the problem is coming from, but I'm suspecting that the problem has always been there (before this PR). Maybe this PR just amplified it so that it became visible.

lyubomir:
I and @gpolitis looked into the SRTP replay attacks and couldn't tie them to the RTP timestamp rewriting.

PR #95, opened 2016-03-02
Adds another method for getting the current sending bitrate for a Med…

bgrozev:
…iaStream.

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamStatsImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamStatsImpl.java
@@ -1548,4 +1548,51 @@ public class MediaStreamStatsImpl
             }
         }
     }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This method is different from {@link #getUploadRateKiloBitPerSec()} in
+     * that:
+     * 1. It is not necessary for {@link #updateStats()} to be called
+     * periodically by the user of libjitsi in order for it to return correct
+     * values.
+     * 2. The returned value is based on the average bitrate over a fixed
+     * window, as opposed to an EWMA.
+     * 3. The measurement is performed after the {@link MediaStream}'s
+     * transformations, notably after simulcast layers are dropped (i.e. closer
+     * to the network interface).
+     *
+     * The return value includes RTP payload and RTP headers, as well as RTCP.
+     */
+    @Override
+    public long getSendingBitrate()
+    {
+        long sbr = -1;
+        AbstractRTPConnector rtpConnector = mediaStreamImpl.getRTPConnector();
+
+        if (rtpConnector != null)
+        {
+            try
+            {
+                RTPConnectorOutputStream rtpStream
+                        = rtpConnector.getDataOutputStream(false);
+                RTPConnectorOutputStream rtcpStream
+                        = rtpConnector.getControlOutputStream(false);
+
+                if (rtpStream != null && rtcpStream != null)
+                {
+                    long now = System.currentTimeMillis();
+                    sbr = rtpStream.getOutputBitrate(now);
+                    sbr += rtcpStream.getOutputBitrate(now);
+                }
+            }
+            catch (IOException ioe)
+            {
+                logger.warn("Failed to get sending bitrate: ", ioe);
+            }
+        }
+
+        return sbr;
+    }
 }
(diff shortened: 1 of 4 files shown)


src/org/jitsi/impl/neomedia/MediaStreamStatsImpl.java
```diff
@@ -1548,4 +1548,51 @@ public void addRembListener(REMBListener listener)
             }
         }
     }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This method is different from {@link #getUploadRateKiloBitPerSec()} in
```
lyubomir: I'll be commenting on the code inline. I consider these thinking-out-loud pieces of thoughts, not issues to be addressed.

I find the method name getUploadRateKiloBitPerSec... unappealing. Is it correct to have bit in there instead of bits. On top of that, its javadoc talks about download.

src/org/jitsi/impl/neomedia/MediaStreamStatsImpl.java
```diff
@@ -1548,4 +1548,51 @@ public void addRembListener(REMBListener listener)
...
+        if (rtpConnector != null)
+        {
+            try
+            {
+                RTPConnectorOutputStream rtpStream
+                        = rtpConnector.getDataOutputStream(false);
+                RTPConnectorOutputStream rtcpStream
+                        = rtpConnector.getControlOutputStream(false);
+
+                if (rtpStream != null && rtcpStream != null)
```
lyubomir: I didn't get it why both RTP and RTCP streams are a must. I guess it could be a requirement of RateStatistics but didn't really search for the cause of the requirement.

src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java
```diff
@@ -123,6 +140,12 @@
 
         PACKET_QUEUE_CAPACITY
             = packetQueueCapacity >= 0 ? packetQueueCapacity : 256;
+
+        logger.debug("Initialized configuration. "
```
lyubomir: We usually do the logging of constructed strings of logger.isDebugEnabled(). I suppose it doesn't matter too much anyway.

src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java
```diff
@@ -215,6 +238,12 @@ static boolean logPacket(long numOfPacket)
     private boolean closed = false;
 
     /**
+     * The instance used to calculate the sending bitrate of this output stream.
+     */
+    private RateStatistics rateStatistics
+        = new RateStatistics(AVERAGE_BITRATE_WINDOW_MS);
```
lyubomir: I usually go with final here. Don't know whether that makes any difference. I appear to remember reading that final is overused and fields are probably the most useful places to use it... or something like that. I may be wrong.

src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java
```diff
@@ -645,6 +675,7 @@ private boolean write(RawPacket[] pkts)
             {
                 if (success)
                 {
+                    rateStatistics.update(pkt.getLength(), now);
                     if (!send(pkt))
```
lyubomir: I didn't get why rateStatistics.update() is before send(). I trust you know why you have it that way so I went and committed it.

src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java
```diff
@@ -663,6 +694,23 @@ private boolean write(RawPacket[] pkts)
         return success;
     }
 
+    /**
+     * @return the current output bitrate in bits per second.
+     */
+    public long getOutputBitrate()
```
lyubomir: OutputBitrate of OutputStream... I think the Output on Bitrate is obsolete but I suppose getBitrate could be viewed as too sparse, vague, etc.

bgrozev: I think it makes it clearer. It could be measuring the bitrate of the stuff that is written to it, before transformations, possible drops of packets, and before it's actually sent (I am moving the call to update() that you mentioned).

PR #67, opened 2015-11-24
Allow user to specify RSA key generation parameters via configuration.

bickelj:
Adds two configuration options:
    org.jitsi.impl.neomedia.transform.dtls.RSA_KEY_SIZE
and
    org.jitsi.impl.neomedia.transform.dtls.RSA_KEY_SIZE_CERTAINTY

The first is the size in bits of the RSA key to be presented by libjitsi on a DTLS handshake.
The second is the certainty passed to BigNumber constructor.

Either may be set by java system properties (e.g. -D) or by the jitsi configuration service (sip-communicator.properties).

A sample on a jitsi-videobridge server updating the key size to 3072 bit RSA keys:
```$ cat ~/.sip-communicator/sip-communicator.properties
org.jitsi.impl.neomedia.transform.dtls.RSA_KEY_SIZE_CERTAINTY=80
org.jitsi.impl.neomedia.transform.dtls.RSA_KEY_SIZE=3072```

Defaults of 1024 and 80, respectively, remain the same as before. Some minor refactoring to move the numbers out of the method doing the generation.

PR #98, opened 2016-03-04
Generate RSA keypair / certificate only once in DtlsControlImpl

champtar:
security implications need to be reviewed carrefully!!
I don't see why we generate a new RSA keypair and certificate for each dtls conn when on a web server we change them every year or more. Is it ok to share the cert for all connections ?

Top 10 functions is now (this includes some unmerged patch):
(t  6,0,s  6,0) java.lang.Object.wait
(t  5,0,s  5,0) java.net.PlainDatagramSocketImpl.send
(t  3,1,s  3,0) java.net.PlainDatagramSocketImpl.receive0
(t  1,5,s  1,5) org.jitsi.impl.neomedia.transform.srtp.OpenSSLHMAC.HMAC_Update
(t  2,5,s  1,1) java.util.concurrent.locks.AbstractQueuedSynchronizer.release
(t  1,7,s  1,1) org.ice4j.socket.MultiplexingXXXSocketSupport.acceptBySocketsOrThis
(t  1,0,s  1,0) java.lang.Object.notifyAll
(t  1,0,s  1,0) java.util.concurrent.locks.ReentrantLock$Sync.tryRelease
(t  1,0,s  1,0) sun.misc.Unsafe.park
(t  1,0,s  1,0) org.jitsi.impl.neomedia.transform.srtp.SRTPCipherCTROpenSSL.AES128CTR_CTX_process

this was:
(t  6,6,s  6,6) java.math.BigInteger.montReduce
(t  5,0,s  3,7) java.math.BigInteger.squareToLen
(t  2,4,s  2,3) java.lang.Object.wait
(t  2,1,s  2,1) java.net.PlainDatagramSocketImpl.send
(t  1,4,s  1,4) java.net.PlainDatagramSocketImpl.receive0
(t  1,9,s  1,3) java.lang.ClassLoader.defineClass1
(t  1,3,s  1,3) java.math.BigInteger.addOne
(t  1,3,s  1,3) java.math.MutableBigInteger.divWord
(t  0,6,s  0,6) java.math.MutableBigInteger.divideOneWord
(t  0,6,s  0,6) org.jitsi.impl.neomedia.transform.srtp.OpenSSLHMAC.HMAC_Update

### DIFF
--- a/src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
+++ b/src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
@@ -248,28 +248,14 @@ public class DtlsControlImpl
                 /**
                  * {@inheritDoc}
                  *
-                 * Employs <tt>ZrtpFortuna</tt> as is common in neomedia. Most
-                 * importantly though, works around a possible hang on Linux
-                 * when reading from <tt>/dev/random</tt>.
+                 * force usage of non blocking pool (/dev/urandom)
                  */
                 @Override
                 public byte[] generateSeed(int numBytes)
                 {
                     byte[] seed = new byte[numBytes];
-
-                    ZrtpFortuna.getInstance().nextBytes(seed);
+                    this.nextBytes(seed);
                     return seed;
-                }
-
-                /**
-                 * {@inheritDoc}
-                 *
-                 * Employs <tt>ZrtpFortuna</tt> as is common in neomedia.
-                 */
-                @Override
-                public void nextBytes(byte[] bytes)
-                {
-                    ZrtpFortuna.getInstance().nextBytes(bytes);
                 }
             };
     }


bickelj:
Downvote.

The purpose of using Jitsi is to have a user-respecting, secure communication tool. The purpose of the server is to serve the user, who is seeking secure communication.

We need strong keys that are rotated frequently. Libjitsi currently does this, and following the RFCs mentioned in the text of the file changed.

There are practical considerations but those are secondary to the premise of weakening security.

bgrozev:
> Downvote.
> 
> The purpose of using Jitsi is to have a user-respecting, secure communication tool. The purpose of the server is to serve the user, who is seeking secure communication.

Can you please elaborate on how using a single certificate per videobridge run would compromise the security? This is an honest question, I'm trying to understand it. 

My very vague understanding is that if DTLS is used in one of the modes which provide perfect forward secrecy, a compromised certificate cannot be used to retroactively obtain any of the sessions keys. And if that's the case, generating a new certificate for each session doesn't seem to increase security in any way. I hesitate to state this, even with the caveat above, so please don't quote me, and please correct me if I am wrong.

bickelj:
> if DTLS is used in one of the modes which provide perfect forward secrecy, a compromised certificate cannot be used to retroactively obtain any of the sessions keys

The thought was that frequent key rotation would be better than infrequent, and any moves to go to less frequent to save a little server power would be moving backwards. But if forward secrecy is guaranteed to be used, maybe it is not a big deal.

champtar:
I just rechecked the code and the cert is generated for 6days only, so we can't merge this 'fix' (will rework on it on Monday).
I agree we should rotate key often, but it would be daily for me, not per connection.

For the strong key part, we are only using 1024 bits key, and we are not using SecureRandom().
Fortuna might be a good CSPRNG, but i'm not sure our implementation as been reviewed/audited, and I really prefer to only use code written by cryptographers and often audited like Bouncy Castle.
RSAKeyGenerationParameters needs a SecureRandom as second argument and Bouncy Castle doesn't provide any SecureRandom implementation, which for me means Bouncy Castle author want java.security.SecureRandom.

If we regenerate keypair daily, blocking SecureRandom() will not be a problem, and recent java (jdk8) use /dev/urandom, so no more blocking.
A really good reading (many links included):
https://tersesystems.com/2015/12/17/the-right-way-to-use-securerandom/

Also DTLS supports PFS cipher suite, but i'm not sure we are enforcing only these suites (haven't checked)

To conclude i will rework on this patch on monday
Regards

bickelj:
Upvote @champtar previous comment.

Please also let us know your thoughts on https://github.com/jitsi/libjitsi/pull/67

champtar:
Hi,
Just updated this PR, I'm still not sure it's ready to merge, ie the fortuna change might need some discussion, but i'm on vacation starting now for one week :)

bgrozev:
@champtar is the change to /dev/urandom independent of the certificate generation? If so, could you please split it in its own PR?

emcho:
This looks good to me! I do think that the 24h rotating should address any practical security concerns.
Thanks very much @champtar ! 
Let's get this in.

gpolitis:
Also, I don't think we're enforcing cipher suites with PFS. If we accept this, we should at least make sure we enforce that.

gpolitis:
My previous message was destined to @champtar, I'm sorry for not mentioning you in the first place.

In summary, if we only support PFS suites, and we use an audited/reviewed SecureRandom implementation and we regenerate the RSA key daily, then even with an RSA key of 1024bits, we should be fine.

champtar:
For now Fortuna is initialy seeded with Random instead of SecureRandom ... and I haven't reviewed the constant seeding (hope it work on VM)

ibauersachs:
If Fortuna is removed here, then why extend SecureRandom at all?

champtar:
generateSeed calls /dev/random, nextBytes calls /dev/urandom, I'm extending SecueRandom to force the use of /dev/urandom to keep old non blocking behaviour. With more testing/review we can remove this hack or add a configuration.
Once kernel pool is initialized, using /dev/urandom is as safe as /dev/random, but we have no way too know that the pool has been initialized (see also kernel getrandom())

champtar:
BTW i've checked SecureRandom behaviour with java 6 7 8 on a CentOS 6 VM

ibauersachs:
It's been a while since I looked at SecureRandom's source, but I have something in mind that it calls getSeed anyway if it hasn't been seeded before. Anyway, there is a (system) property to force the usage uf urandom (by setting e.g. /dev/../dev/urandom to trick Java). Can't we set that in jvb.sh?

bickelj:
@ibauersachs https://bugs.openjdk.java.net/browse/JDK-6202721

@gpolitis If we're worried about the length of time it takes to generate a key and therefore reducing key generation to once a day instead of every connection, why not make it stronger than 1024 bits?

lyubomir:
I'm working on merging https://github.com/champtar/libjitsi/commit/e13d0471a5c11ec187cf32c02289a46d82811bdd now because I need it to speed up the connection time between Meet and Videobridge.

champtar:
Hi @lyubomir,
Still on vacation till wednesday, just merge the RSA cache commit (or another version, i really don't mind the ownership) and i will open another PR with the fortuna removal commit, and another PR to enforce PFS if you don't do it

lyubomir:
I cherry-picked https://github.com/champtar/libjitsi/commit/e13d0471a5c11ec187cf32c02289a46d82811bdd (i.e. Cache RSA keypair for 24h in DtlsControlImpl) because (as I said above) I need it for execution time improvements. Authorship is preserved (I believe).

About ZrtpFortuna (i.e. https://github.com/champtar/libjitsi/commit/a49b2f5326b64865353eea0e2584d4e67b96bb68), why do we have to modify DtlsControlImpl? I thought we made conscious effort to use ZrtpFortuna throughout our source code (admittedly, I'm not sure that we did it everywhere). Can't we fix ZrptFortunaEntropyGatherer  instead?

champtar:
I'm no cryto expert but i can tell you that fortuna as implemented might not be secure for now (it's initialized by Random instead of SecureRandom). Why do you need something else than /dev/urandom? Who made that choice? For what reasons?

ibauersachs:
Where is Fortuna seeded by `Random`?

ibauersachs:
@lyubomir I'm not sure `ZrptFortunaEntropyGatherer`works as expected if run on a server or VM. There will be no audio device and it might not seed Fortuna at all.

champtar:
@ibauersachs I'm talking about initial seeding https://github.com/wernerd/ZRTP4J/blob/master/src/gnu/java/zrtp/utils/ZrtpFortuna.java#L48

The only way to fix this is to remove it.
On a VM there is no source of entropy that why you install special software/device to add some.
Please read the links i've put in the discussion, they include the links to the Android bitcoin wallets fail (prng was not properly seeded)

lyubomir:
@ibauersachs, by fixing ZrtpFortunaEntropyGatherer I meant to add something else in addition to the audio capture device (which is not available on the server, of course). Can't we do something with /dev/urandom there? I don't feel awfully excited about doing something with /dev/urandom in DtlsControlImpl and not doing the same elsewhere where we have ZrtpFortuna.

ibauersachs:
Can't we just use `SecureRandom` in the fortuna gatherer and force it to use urandom in jvb.sh/bat by setting `-Djava.security.egd=file:/dev/./urandom`?

champtar:
Nobody answered why do we need Fortuna in the first place?

lyubomir:
@champtar, I don't know whether we need Fortuna. I guess it's being used by the ZRTP-related functionality in libjitsi or the third-party library but I don't have the time to check. Fortuna is a singleton to something may be using it even if we don't see it in libjitsi's source code.

ibauersachs:
zrtp4j needs it (seeded). So we can't just drop the seeding of Fortuna from libjitsi entirely, but the current state is unsafe.

lyubomir:
@ibauersachs, thank you! I absolutely agree with you that we need to fix an unsafe situation, I'm not debating that there's an unsafe situation. I merely want to make sure that we're not fixing it in DTLS only and leaving the rest of the Jitsi universe with the same issue.

ibauersachs:
Hm, thinking a bit more about it I'm not sure we can even fix this outside of the zrtp4j library. The singleton is used from an enum-constructor. It's possible that this already causes Fortuna to initialize with bad data, especially because zrtp4j needs to be loaded by OSGi before the MediaService is started.
Fortuna came from @wernerd, so I don't know why it was used over SecureRandom. But the more I think about it, the more I think we should get completely rid of it.

champtar:
see #108

PR #108, opened 2016-03-17
Cache certificate and fingerprint, stop using ZrtpFortuna

champtar:


ibauersachs:
It ain't as simple as that :(
I know that I had to introduce setRandomGenerator on SDesFactory because it was blocking on /some/ (desktop) machines. We need to guarantee that Java never reads from /dev/random, not even for seeding. If we're using multiple instances of SecureRandom, generateSeed(..) is called multiple times and will drain /dev/random.

Also please split the Fortuna-removal and cache thing into two PRs. About the cache commit: the _ in the variable names are nothing we ever used. Please remove those. I cannot comment on the actual change there.

champtar:
generateSeed is never called, and if you don't use it /dev/random is never called
i've tested with java 6/7/8 and strace, is it possible that it was before java 6?

will update this PR once #109 is merged

champtar:
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/sun/security/provider/NativePRNG.java

champtar:
by the way ZrtpFortuna do use SecureRandom
https://github.com/jitsi/zrtp4j/commit/1bd0743d9cdd1750def56ff60951a99d44db2a87

lyubomir:
I'll look at https://github.com/jitsi/libjitsi/pull/108/commits/584363a4be3549c94005544597e55c9164264b8e because I have similar modifications in my local clone.

However, I'm going to close this PR now because (1) I will not pull Fortuna now and (2) I believe there are two distinct issues here: caching and Fortuna. Please @champtar create a separate PR for Fortuna which deals with Fortuna only (I'll cherry-pick the caching from here now).

champtar:
@lyubomir see #109

PR #111, opened 2016-03-19
Sctp fixes

bgrozev:


paweldomas:
I'd rather leave the decision on using thread pool to the application which can be implemented in this interface here:
https://github.com/jitsi/libjitsi/blob/master/src/org/jitsi/sctp4j/NetworkLink.java#L38

bgrozev:
Absolutely!

PR #113, opened 2016-03-21
Use SecureRandom instead of ZrtpFortuna

champtar:
Drop createSecureRandom()

On linux, new SecureRandom() uses /dev/urandom for nextBytes() calls
and /dev/random for generateSeed().
I've found no call to generateSeed, so no blocking calls.
I've also double checked with "strace -e trace=read -y -f <jvb command>"
and i only see /dev/urandom reads (no /dev/random)
and only for connections initialisation

We are not sure of ZrtpFortuna safety at 100%:
on my desktop dev machine ZrtpFortunaEntropyGatherer.setEntropy()
return false (audioCaptureDevice == null) which mean that no reseeding take place
Initial seeding is made with SecureRandom since
https://github.com/jitsi/zrtp4j/commit/1bd0743d9cdd1750def56ff60951a99d44db2a87

Profiling show no performance impact, so better use a know good CSPRNG

P.S:
some discussion already took place in #108 and #98


champtar:
tests also ok here, is jenkins broken?

damencho:
We still haven't done publishing the logs and results to public. But I see 9 tests fail, will check it later can it be a problem with the environment ... these are the failing:
 org.jitsi.meet.test.AvatarTest.changeAvatarAndCheck
 org.jitsi.meet.test.AvatarTest.avatarWhenVideoMuted
 org.jitsi.meet.test.StopVideoTest.startVideoOnParticipantAndCheck
 org.jitsi.meet.test.SwitchVideoTests.participantClickOnLocalVideoAndTest
 org.jitsi.meet.test.ActiveSpeakerTest.testActiveSpeaker
 org.jitsi.meet.test.StartMutedTest.checkboxesTest
 org.jitsi.meet.test.StartMutedTest.configOptionsTest
 org.jitsi.meet.test.DisplayNameTest.testChangingDisplayName
 org.jitsi.meet.test.ContactListTest.testContactList

jitsi-developers:
All tests are failing, as participants are starting muted for some reason
... Working on that now.

damencho:
Jenkins: it's ok to test.

lyubomir:
@champtar, I believe I shared in one of your previous PRs (with @ibauersachs if memory serves me), that I'd rather we take a global decision for Fortuna, not locally to DtlsControlImpl. I'd rather we don't merge this PR (in its current state in which it deals with Fortuna in DtlsControlImpl only).

champtar:
I can re-add the removal for the rest of libjitsi, it's just not tested (zrtp and sdes), but really simple. I've extensively tested this commit and I'm sure there are no calls to /dev/random. I'm in favour of dropping Fortuna and willing to contribute all needed patches to nuke it.

lyubomir:
Cool. Let's hear from @ibauersachs as well before we move forward.

ibauersachs:
strace is not conclusive if you only use it in Videobridge. I know that Java calls or used to call /dev/random [1] and [2]. I need to find out where (browsing Java source on GrepCode isn't my favorite thing to do at midnight) before we remove Fortuna in ZRTP and SDES. I don't mind in DTLS, as this is only used in the Videobridge anyway.

[1] http://lists.jitsi.org/pipermail/dev/2011-October/004518.html
[2] https://docs.oracle.com/javase/7/docs/api/java/security/SecureRandom.html (see the note)

champtar:
hi all,
i've re-read https://tersesystems.com/2015/12/17/the-right-way-to-use-securerandom/
depending on the content of java.security, some distro might use SHA1PRNG calling /dev/random to init itself.

As we are already using one SecureRandom for ZRTP seeding (https://github.com/jitsi/zrtp4j/commit/1bd0743d9cdd1750def56ff60951a99d44db2a87), what do you think about a SecureRandom singleton?

org.jitsi.util.SecureRandomProvider.getInstance()

ibauersachs:
As far as I could see yesterday, the seed of SecureRandom is already a singleton. But I need to investigate that further.

champtar:
some more testing:
```
import java.security.SecureRandom;

public class SecureRandomTests {

    public static void main(String [ ] args) throws Exception
    {
        while (true) {
            //SecureRandom random = new SecureRandom();
            SecureRandom random = SecureRandom.getInstance("NativePRNG");
            //SecureRandom random = SecureRandom.getInstance("SHA1PRNG");
            //random.generateSeed(40);
            byte bytes[] = new byte[1000];
            random.nextBytes(bytes);
            System.out.println("random\n");
            Thread.sleep(1000);
        }
    }
}
```

calling ```SecureRandom.getInstance("SHA1PRNG").nextBytes()``` multiple times only read 20 bits from ```/dev/random``` on first use (seeding), and never after.

calling ```SecureRandom.getInstance("NativePRNG").nextBytes()``` always read from ```/dev/urandom``` (same as with ```new SecureRandom()```)

calling ```generateSeed()``` always read from ```/dev/random``` (even for ```SHA1PRNG```)

Tested on centos 6 java 6/7/8, we need some testing on mac/win

champtar:
@ibauersachs is it possible that your bug report from 2011 was with a prehistoric version like java 1.4?

ibauersachs:
I can't tell the Java version anymore, but definitely >= 1.5, most likely 1.6. Windows is irrelevant, there is no /dev/random (they either don't use native at all there or it is CAPI, which is never slow).

champtar:
Do you remember if it was linux or mac?

ibauersachs:
I tested on Ubuntu, most likely 11.04. Stephane, who reported it initially [1], was on 10.04.

[1] http://lists.jitsi.org/pipermail/dev/2011-October/004524.html

champtar:
hi,
just tested with ubuntu 10.04.4 (http://old-releases.ubuntu.com/releases/10.04.3/ubuntu-10.04.4-desktop-amd64.iso) and openjdk-6-jdk, same behaviour, ie ```SHA1PRNG``` read 20 bits of /dev/random only on first use (seeding), ```NativePRNG``` calls /dev/urandom

ZRTPFortuna in SDES seems to be there from day one (150fd97)

Is it easy to test SDES and ZRTP negotiation?
i can re-add my 2 patchs and we just test if it work or not

ibauersachs:
ZrtpFortuna came in around SVN rev ~8500, i.e. before libjitsi was taken out of Jitsi (desktop). You could test ZRTP and SDES by installing Jitsi on two computers that aren't firewalled/natted between each other and set up a server-less SIP account on both.

champtar:
i've never built jitsi, is there an howto somewhere

champtar:
i've re-added missing commits

ibauersachs:
```
git clone https://github.com/jitsi/jitsi
cd jitsi
ant make
```
Or download the .deb packages.

ibauersachs:
I was just looking at sdes4j's source and found that I've been using `generateSeed` before I introduced `setRandomGenerator` - which was obviously wrong [1].

https://github.com/ibauersachs/sdes4j/commit/987a617670924894d48b105f0a291c88107bb47a

champtar:
ok, so we can drop Fortuna now!

champtar:
Ping

champtar:
just rebased (there was a conflict)

champtar:
jenkins broken again?

damencho:
No, there are currently various problem in meet, in lib-jitsi-meet and in jvb and most of the tests are failing, and we are working on it. Its an actual problem.

ibauersachs:
So, finally went through the sources of Java to understand what's going on.

Java 6, 7:
nativeAvail = exists(/dev/random) && exists(/dev/urandom)
if (nativeAvail && securerandom.source==file:/dev/urandom) -> new SecureRandom() == NativePRNG

Java 8:
nativeAvail = canread(/dev/random) && canread(/dev/urandom)
randSource = (securerandom.source==file:/dev/urandom || securerandom.source==file:/dev/random)
if (nativAvail && ) -> new SecureRandom() == NativePRNG

(On Windows ::CryptGenRandom() is used).

/dev/random is used to seed, then /dev/urandom for nextBytes(...). If /dev/random is drained while seeding 20 bytes, `new SecureRandom()` can block. Based on this, can you please extend this PR with:
- Drop org.jitsi.impl.neomedia.transform.zrtp.ZrtpFortunaEntropyGatherer completely
- Replace contents MediaServiceImpl.postInitializeOnce with
```
        logger.info("Seeding SecureRandom...");
        SecureRandom rnd = new SecureRandom();
        byte[] b = new byte[20];
        rnd.nextBytes(b);
        logger.info("SecureRandom seeding finished. Bytes read: " + b.length);
```

If /dev/random is drained, we block during startup. But it will guarantee us that we won't block while the application (JVB, Jitsi) is running.

champtar:
zrtp4j still uses Fortuna, that's why i'm not dropping ZrtpFortunaEntropyGatherer right now.
I can provide patches for zrtp4j if you want.

ibauersachs:
Fortuna in zrtp4j is seeded with SecureRandom, so this doesn't matter. Or what did I miss here?

champtar:
I'm speaking from memory, i'll recheck

champtar:
@ibauersachs i've added needed commits

src/org/jitsi/impl/neomedia/MediaServiceImpl.java
```diff
@@ -1531,9 +1532,17 @@ public static boolean isJmfRegistryDisableLoad()
...
+         * Some SecureRandom() implementations like SHA1PRNG call
+         * /dev/random to seed themselves on first use.
+         * Call SecureRandom early to avoid blocking when establishing
+         * a connection for exemple.
+         */
+        logger.info("Warming up SecureRandom...");
+        SecureRandom rnd = new SecureRandom();
+        byte[] b = new byte[20];
+        rnd.nextBytes(b);
+        logger.info("Warming up SecureRandom finished.");
```
ibauersachs: Do we have a guarantee that this code isn't thrown away by an optimizing compiler? That's why I printed the byte-array length (as useless as the information actually is).

champtar: I don't see how a sane compiler can decide to remove a function call like ```rnd.nextBytes(b);```.
how can it decide than there is no side affect when there actually are (SHA1PRNG seeding or NativePrng file opening)
Unless you know a buggy java version doing such harmful optimisation, i'm against defensive coding.

ibauersachs: Object on stack doing lots of stuff using reflection, I can imagine a compiler seeing this as 'does nothing, throw away'. Anyway, javac doesn't do much optimization anyway and leaves to Hotspot. So I guess it should be safe enough.

bgrozev:
This causes major problems for videobridge. The DTLS client uses SecureRandom#generateSeed which ends up blocking (on /dev/random, I suppose) for a long time (over 50sec on one of the AWS VMs which we regularly use).

    at org.jitsi.impl.neomedia.transform.dtls.DtlsPacketTransformer$SecureRandom2.generateSeed(DtlsPacketTransformer.java:1756)
    at org.bouncycastle.crypto.tls.AbstractTlsContext.<init>(Unknown Source)
    at org.bouncycastle.crypto.tls.TlsClientContextImpl.<init>(Unknown Source)
    at org.bouncycastle.crypto.tls.DTLSClientProtocol.connect(Unknown Source)
    at org.jitsi.impl.neomedia.transform.dtls.DtlsPacketTransformer.runInConnectThread(DtlsPacketTransformer.java:1011)
    at org.jitsi.impl.neomedia.transform.dtls.DtlsPacketTransformer.access$100(DtlsPacketTransformer.java:40)
    at org.jitsi.impl.neomedia.transform.dtls.DtlsPacketTransformer$2.run(DtlsPacketTransformer.java:1281)

champtar:
I don't find generateSeed in libjitsi or bouncycastle 1.54, I'll look at it tomorrow

bgrozev:
I see this with openjdk 1.7.0_79 on ubuntu 14.04. It is not only on initialization (or the first call), and I do see it reading from /dev/random.

bgrozev:
This is possibly because of the BC version: jitsi-videobridge master uses 1.51-jitsi-1 which doesn't include this: https://github.com/bcgit/bc-java/commit/6bbae6fec39f0940a0f3a1bcc96a062da306db6f

ibauersachs:
@bgrozev this should be gone when https://github.com/jitsi/libjitsi/pull/128 is merged.

champtar:
sorry for this one
While doing this PR i've checked the source code to be sure that there is no generateSeed calls, but i've only checked BC 1.54, not 1.51.
@bgrozev now that BC is updated, can you re-test

bgrozev:
OK, thanks for the quick reaction. I'll write to confirm once I test (but I expect it doesn't occur anymore, so I won't be running a test specifically for that).

bgrozev:
I didn't see any issues after testing.

PR #115, opened 2016-03-23
Packet queue

bgrozev:


pom.xml
```diff
@@ -55,7 +55,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>ice4j</artifactId>
-      <version>1.0-20160310.155819-18</version>
+      <version>1.0-SNAPSHOT</version>
```
brianh5: this was probably just for dev?

bgrozev: Yes. The file has to be updated with the new snapshot version before merging (should have mentioned that in the PR comment).

src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java
```diff
@@ -768,11 +768,19 @@ public long getOutputBitrate(long now)
          */
         final Thread sendThread;
 
+        QueueStatistics queueStats = null;
```
paweldomas: missing JavaDocs ? also why does it have "package private" access level ? Just asking...

bgrozev: Added javadoc. I didn't use the "private" keyword, because these are members of a private inner class, so they cannot be accessed from the outside (unless I am missing something).

PR #119, opened 2016-03-26
Unbreak DTLS flights

lyubomir:
The following modifications were applied while trying to remove a 1 second hard minimum on connectivity establishment between Videobridge and Chrome/WebRTC. Related modifications in jitsi-videobridge (i.e. which will be introduced in a branch there with the same name) (will) depend on these.

Of separate interest is commit 32c3e34 which fixes a long-standing issue of breaking DTLS flights into multiple UDP packets. The fix is desirable in the context of optimizing the connectivity establishment i.e. this PR.

### DIFF
--- a/src/org/jitsi/impl/neomedia/transform/dtls/DatagramTransportImpl.java
+++ b/src/org/jitsi/impl/neomedia/transform/dtls/DatagramTransportImpl.java
@@ -509,8 +509,12 @@ public class DatagramTransportImpl
                 case HandshakeType.hello_request:
                 case HandshakeType.hello_verify_request:
                 case HandshakeType.server_hello_done:
+                    endOfFlight = true;
+                    break;
                 default:
                     endOfFlight = true;
+                    logger.warn(
+                            "Unknown DTLS handshake message type: " + msg_type);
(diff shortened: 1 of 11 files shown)


src/org/jitsi/impl/neomedia/transform/dtls/DatagramTransportImpl.java
```diff
@@ -482,7 +482,15 @@ public void send(byte[] buf, int off, int len)
...
-                short msg_type = TlsUtils.readUint8(buf, off + 11);
+                short msg_type
+                    = TlsUtils.readUint8(
+                            buf,
+                            off
+                                + 1 /* type */
+                                + 2 /* version */
+                                + 2 /* epoch */
+                                + 6 /* sequence_number */
+                                + 2 /* length */);
```
bgrozev: Maybe we should print a warning when we encounter an unknown message type (the "default" in the switch statement below)?

lyubomir: OK, thank you. I'll do that.

PR #129, opened 2016-04-06
Fixes and improvements for RTX.

bgrozev:
Recognizes packets retransmitted with RTX and doesn't request them again.

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -342,11 +342,12 @@ public class MediaStreamImpl
     private TransformEngine transformEngineChain;
 
     /**
-     * The {@code RetransmissionRequester} instance for this
+     * The {@code RetransmissionRequesterImpl} instance for this
      * {@code MediaStream} which will request missing packets by sending
      * RTCP NACKs.
      */
-    private RetransmissionRequester retransmissionRequester;
+    private final RetransmissionRequesterImpl retransmissionRequester
+        = createRetransmissionRequester();
 
     /**
      * Initializes a new <tt>MediaStreamImpl</tt> instance which will use the
@@ -972,10 +973,11 @@ public class MediaStreamImpl
     }
 
     /**
-     * Creates the {@link RetransmissionRequester} for this {@code MediaStream}.
-     * @return the created {@link RetransmissionRequester}.
+     * Creates the {@link RetransmissionRequesterImpl} for this
+     * {@code MediaStream}.
+     * @return the created {@link RetransmissionRequesterImpl}.
      */
-    protected RetransmissionRequester createRetransmissionRequester()
+    protected RetransmissionRequesterImpl createRetransmissionRequester()
     {
         return null;
     }
@@ -1034,7 +1036,6 @@ public class MediaStreamImpl
             statisticsEngine = new StatisticsEngine(this);
         engineChain.add(statisticsEngine);
 
-        retransmissionRequester = createRetransmissionRequester();
         if (retransmissionRequester != null)
         {
             engineChain.add(retransmissionRequester);
@@ -3631,5 +3632,13 @@ public class MediaStreamImpl
     public RawPacketCache getPacketCache()
     {
         return cachingTransformer;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public RetransmissionRequester getRetransmissionRequester()
+    {
+        return retransmissionRequester;
     }
 }
--- a/src/org/jitsi/impl/neomedia/RawPacket.java
+++ b/src/org/jitsi/impl/neomedia/RawPacket.java
@@ -1365,9 +1365,9 @@ public class RawPacket
      *
      * @return the OSN value of an RTX packet.
      */
-    public short getOriginalSequenceNumber()
+    public int getOriginalSequenceNumber()
     {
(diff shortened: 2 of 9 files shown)


brianh5:
hey boris, we've started doing some work on enabling retransmission and have found that h264 requires the rtx stream to be separate (different ssrc, different pt) all the way to the receiver (vp8 can do it in band).  i'm wondering, is there a scenario where the sender says it wants to do retransmissions on a separate stream, but the receiver can handle them on the same stream?  from what we've seen with vp8 and h264 it seems to be one way or another (either both sender and receiver can handle them on the same stream, or they need to be kept separate the entire way).

brianh5:
one other comment: of course if the retransmission is done in time to mask the loss entirely from the downstream receiver, then this would work.  but what if there is loss on the downlink and the receiver requests a retransmission?

brianh5:
is the idea that the server would request rtx for loss on the uplink and mask them to look like the original stream, but still negotiate rtx with the receiver and, when the receiver reported loss, would retransmit them on the separate ssrc/payload type?

bgrozev:
Hey Brian,

Withour RTX webrtc.org does retransmissions "plain" -- by just sending the original packet again. The RTX format is the one which uses a separate stream (ssrc, pt, sequence numbers). I think that both can be used with h264, but there might be limitations in webrtc.org's h264 implementation that I'm not aware of.

Connecting endpoints which support RTX with ones that don't support it is a use case we built for, because firefox doesn't support RTX. The way this works is that the bridge terminates retransmissions.

With a sender: the bridge will request missing packets (with RTCP NACK). They may arrive either as copies of the lost packet or in RTX. If they arrive as RTX, the bridge will strip it and pass on a regular media packet.

With a receiver: the bridge will terminate incoming RTCP NACK packets. It will try to respond with a packet from its cache, with either RTX or a plain copy, depending on the receiver. If the packet is not in the cache, the bridge will do nothing: presumably the bridge already detected the loss and requested a retransmission from the sender, which will find its way to the receiver as a media packet (no RTX).

Does that answer your questions?

champtar:
@gpolitis why have you merged code that failed to pass the tests ?
Tests don't even compile for me

bgrozev:
@champtar, I'm sorry if this caused problems. It may have been opened and merged a bit too quickly, but I don't see any compilation failures. What exactly fails to compile for you? 

We are looking into issues which might have been caused by this, but we don't know of any that would have an impact unless you enable RTX (which would not have worked prior to these PRs either).

gpolitis:
As @bgrozev said, this might have been merged a little too quickly. @champtar to which tests are you referring to? Unit tests? They compile fine here. The tests that jenkins is running are the jitsi-meet-torture tests but they're not very reliable and failure doesn't necessarily mean that something is wrong (we're working on fixing that, but that's a different issue).

gpolitis:
Actually there are problems with RTX and they have been captured by the tests, so we're temporarily disabling it. I apologize for the quick merge.

champtar:
I will paste compilation bug tomorrow, I'm out of office right now.

champtar:
https://github.com/jitsi/jicofo/commit/178d0a3edb7924f37ed0d54ea7ce920a503e034a

bgrozev:
Got it, thanks @champtar ! I think Pawel fixed this already. I wasn't seeing it in my local env for some reason (running "mvn install"). Apologies for the whole mess, I think it is all in a working state now, but we'll be testing more thoroughly today.

brianh5:
Thanks Boris!  I think so, let me see if I'm understanding it right.

> Withour RTX webrtc.org does retransmissions "plain" -- by just sending the original packet again. The RTX format is the one which uses a separate stream (ssrc, pt, sequence numbers). I think that both can be used with h264, but there might be limitations in webrtc.org's h264 implementation that I'm not aware of.

We're still trying to figure out which configurations work and which don't work for h264 at this point.  Will keep you posted on what we find there.

> With a sender: the bridge will request missing packets (with RTCP NACK). They may arrive either as copies of the lost packet or in RTX. If they arrive as RTX, the bridge will strip it and pass on a regular media packet.

Makes sense and we've seen this.  So on the far end it will just look like an out-of-order packet and be handled by the jitter buffer I assume?


> With a receiver: the bridge will terminate incoming RTCP NACK packets. It will try to respond with a packet from its cache, with either RTX or a plain copy, depending on the receiver. If the packet is not in the cache, the bridge will do nothing: presumably the bridge already detected the loss and requested a retransmission from the sender, which will find its way to the receiver as a media packet (no RTX).

So does this mean that if a receiver has negotiated rtx and a separate payload type, when it sends a nack the bridge will send the retransmitted packet on the rtx stream using the rtx payload type?

bgrozev:
> With a sender: the bridge will request missing packets (with RTCP NACK). They may arrive either as copies of the lost packet or in RTX. If they arrive as RTX, the bridge will strip it and pass on a regular media packet.
> 
> Makes sense and we've seen this. So on the far end it will just look like an out-of-order packet and be handled by the jitter buffer I assume?

Yes.

> So does this mean that if a receiver has negotiated rtx and a separate payload type, when it sends a nack the bridge will send the retransmitted packet on the rtx stream using the rtx payload type?

Yes, exactly.

champtar:
@bgrozev @gpolitis i'm not using fixed dependencies and always rebuilding all projects, that's why jicofo was not compiling for me. Thanks for the fixes, I'll update tomorrow.

PR #130, opened 2016-04-07
OveruseEstimator replace LinkedList<Double> with double[]

champtar:
this avoid a lot of autoboxing, and is more efficient

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
+++ b/src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
@@ -86,7 +86,17 @@ class OveruseEstimator
 
     private double slope;
 
-    private final List<Double> tsDeltaHist = new LinkedList<>();
+    /**
+     * Store the tsDelta history into a {@code double[]} used as a circular buffer
+     * Original c++ code uses std::list<double> but in Java this translate
+     * to {@code List<Double>} which causes a lot of autoboxing
+     */
+    private final double[] tsDeltaHist = new double[kMinFramePeriodHistoryLength];
+
+    /**
+     * Index to insert next value into {@link tsDeltaHist}
+     */
+    private int tsDeltaHistInsIdx;
 
     private double varNoise;
 
@@ -99,6 +109,11 @@ class OveruseEstimator
         varNoise = options.initialVarNoise;
         E = clone(options.initialE);
         processNoise = options.initialProcessNoise.clone();
+        /**
+         * Initialize {@link tsDeltaHist} with {@code Double.MAX_VALUE}
+         * to simplify {@link updateMinFramePeriod}
+         */
+        Arrays.fill(tsDeltaHist, Double.MAX_VALUE);
     }
 
     /**
@@ -229,14 +244,24 @@ class OveruseEstimator
 
     private double updateMinFramePeriod(double tsDelta)
     {
-        double minFramePeriod = tsDelta;
+        /**
+         * Change from C++ version:
+         * We use {@link tsDeltaHist} as a circular buffer initialized
+         * with {@code Double.MAX_VALUE}, so we insert new {@link tsDelta}
+         * value at {@link tsDeltaHistInsIdx} and we search for the
+         * minimum value in {@link tsDeltaHist}
+         */
+        tsDeltaHist[tsDeltaHistInsIdx] = tsDelta;
+        tsDeltaHistInsIdx = (tsDeltaHistInsIdx + 1) % tsDeltaHist.length;
 
-        if (tsDeltaHist.size() >= kMinFramePeriodHistoryLength)
-            tsDeltaHist.remove(0);
-        for (Double d : tsDeltaHist)
-            minFramePeriod = Math.min(d, minFramePeriod);
-        tsDeltaHist.add(tsDelta);
-        return minFramePeriod;
+        double min = tsDelta;
+        for (double d : tsDeltaHist)
+        {
+            if (d < min)
+                min = d;
+        }
+
+        return min;
     }
 
     private void updateNoiseEstimate(


src/org/jitsi/impl/neomedia/transform/srtp/BaseSRTPCryptoContext.java
```diff
@@ -268,7 +268,7 @@ protected BaseSRTPCryptoContext(
      * @param pkt the RTP packet to be authenticated
      * @param rocIn Roll-Over-Counter
      */
-    synchronized protected void authenticatePacketHMAC(RawPacket pkt, int rocIn)
+    protected void authenticatePacketHMAC(RawPacket pkt, int rocIn)
```
lyubomir: How is this related to the subject of this PR and/or OveruseEstimator? @champtar, could I please ask you to keep PRs to a single, standalone subject? If you need to work on two dependent subjects, you can still create two PRs and state in one of them that it depends on the other and paste a link to it.

lyubomir:
Jenkins, test this please.

src/org/jitsi/impl/neomedia/transform/srtp/BaseSRTPCryptoContext.java
champtar: This should definitly not be here (git commit -a ...), thanks for the catch, I'll remove it

champtar:
I've removed the intruder :)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
```diff
@@ -86,7 +86,8 @@
 
     private double slope;
 
-    private final List<Double> tsDeltaHist = new LinkedList<>();
+    private final double[] tsDeltaHist = new double[kMinFramePeriodHistoryLength];
+    private int tsDeltaHistIdx = 0;
```
lyubomir: As you've probably figured out based on the sparse class comment that I've put, OveruseDetector was translated from C++. Consequently, it lacks javadocs where there no equivalents in the respective C++ source code.

However, we need comments where we author source code. And in your case, I'd say that javadocs are pretty necessary because you're practically implementing a circular buffer and that's complex enough to not be immediately recognizable. Without good javadocs, I'd say the supposed performance improvement is not big enough to justify modifying a translated piece.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
```diff
@@ -99,6 +100,8 @@ public OveruseEstimator(OverUseDetectorOptions options)
         varNoise = options.initialVarNoise;
         E = clone(options.initialE);
         processNoise = options.initialProcessNoise.clone();
+        //init with max value to simplify updateMinFramePeriod
```
lyubomir: I don't like such formatting of comments. Make it a proper sentence in written English please.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
lyubomir: I think the part Idx of the name tsDeltaHistIdx could be more meaningful. Index is generic enough to not bring much but the fact that it's an index (because the fact that it's related to the array tsDeltaHist is represented by the prefix tsDeltaHist). If I'm reading your source code right, you're implementing a circular buffer and tsDeltaHistIdx is the tail (or the insertion head but that's another discussion), Wouldn't it be better to reflect more meaning in the name?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
```diff
@@ -229,14 +232,18 @@ public void update(
...
-        double minFramePeriod = tsDelta;
-
-        if (tsDeltaHist.size() >= kMinFramePeriodHistoryLength)
-            tsDeltaHist.remove(0);
-        for (Double d : tsDeltaHist)
-            minFramePeriod = Math.min(d, minFramePeriod);
-        tsDeltaHist.add(tsDelta);
-        return minFramePeriod;
+        tsDeltaHist[tsDeltaHistIdx] = tsDelta;
+        tsDeltaHistIdx = (tsDeltaHistIdx + 1) % tsDeltaHist.length;
```
lyubomir: Is that (much) faster than the following? I'm curious if you have the answer (i.e. I don't want you to research it or change your source code).
```java
int nextTsDeltaHistIdx = tsDeltaHistIdx + 1;
tsDeltaHistIdx = (nextTsDeltaHistIdx < tsDeltaHist.length) ? nextTsDeltaHistIdx : 0;
```

champtar: i've no idea, it's just how it went out of my brain :)

champtar:
hi @lyubomir 
i've updated the commit with more javadocs

champtar:
here is the top 20 functions while profiling JVB:
```
	(t  9,2,s  9,1) java.lang.Object.wait
	(t  5,8,s  5,8) java.net.PlainDatagramSocketImpl.send
	(t  4,9,s  4,9) java.net.PlainDatagramSocketImpl.receive0
	(t  2,7,s  2,7) sun.misc.Unsafe.park
	(t  3,0,s  2,3) java.util.concurrent.locks.AbstractQueuedSynchronizer.release
	(t  2,0,s  2,0) org.jitsi.impl.neomedia.transform.srtp.OpenSSLHMAC.HMAC_Update
	(t  1,7,s  1,7) java.lang.Object.notifyAll
	(t  1,7,s  1,6) java.util.concurrent.ArrayBlockingQueue.enqueue
	(t  1,5,s  1,5) org.jitsi.impl.neomedia.transform.srtp.SRTPCipherCTROpenSSL.AES128CTR_CTX_process
	(t  2,2,s  1,4) org.ice4j.socket.MultiplexingXXXSocketSupport.acceptBySocketsOrThis
	(t  1,3,s  1,3) sun.misc.Unsafe.unpark
	(t  1,3,s  1,0) org.jitsi.impl.neomedia.RTPConnectorInputStream.read
	(t  1,0,s  1,0) java.util.HashMap.getNode
	(t  1,0,s  1,0) java.lang.Double.valueOf
	(t  0,8,s  0,8) org.jitsi.impl.neomedia.transform.srtp.OpenSSLHMAC.HMAC_Final
	(t  8,4,s  0,7) org.ice4j.socket.MultiplexingXXXSocketSupport.receive
	(t  0,7,s  0,7) java.lang.Thread.setDaemon
	(t  0,6,s  0,6) java.lang.Object.notify
	(t  8,8,s  0,6) org.ice4j.ice.harvest.SinglePortUdpHarvester.runInHarvesterThread
	(t  0,6,s  0,6) java.util.concurrent.ConcurrentHashMap.get
```

```java.lang.Double.valueOf``` is 14th and using 2/3 the time used by ```AES128CTR_CTX_process``` (1,0 vs 1,5) so this isn't a major performance improvements but still is good to take

PR #97, opened 2016-03-03
Rework encryption / openssl offloading

champtar:
this need a rebuild of jnopenssl

PR #133, opened 2016-04-08
Remove reference to NIOBlockCipher in CryptoBenchmark

### LINKED ISSUE #132: Cannot build libjitsi: CryptoBenchmark still references deleted NIOBlockCipher class
Maven build failed due to unresolved class NIOBlockCipher used in CryptoBenchmark:

```
NIOBlockCipher nioBlockCipher
                    = (blockCipher instanceof NIOBlockCipher)
                        ? (NIOBlockCipher) blockCipher
                        : null;
```

[CryptoBenchmark.java:163](https://github.com/jitsi/libjitsi/blob/master/src/org/jitsi/impl/neomedia/transform/srtp/CryptoBenchmark.java#L163)

Class was deleted month ago: https://github.com/jitsi/libjitsi/commit/18c951646db8d748f17d6aaeb7feec37439266f9#diff-2d6e07f58eb241e2530d748450b72b6a

champtar:
NIOBlockCipher as been removed in PR #97
commit 18c951646db8d748f17d6aaeb7feec37439266f9

This was missed because CryptoBenchmark.java was excluded
in maven compilation, so don't exclude it

This fixes #132 (Broken Ant build)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,11 +83,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
-          <excludes>
-            <exclude>org/jitsi/impl/neomedia/transform/srtp/CryptoBenchmark.java</exclude>
-          </excludes>
-        </configuration>
       </plugin>
       <plugin>
          <groupId>org.apache.felix</groupId>
--- a/src/org/jitsi/impl/neomedia/transform/srtp/CryptoBenchmark.java
+++ b/src/org/jitsi/impl/neomedia/transform/srtp/CryptoBenchmark.java
@@ -34,14 +34,11 @@ public class CryptoBenchmark
         throws Exception
     {
         boolean benchmarkJavaxCryptoCipher = false;
-        boolean benchmarkNIOBlockCipher = false;
 
         for (String arg : args)
         {
             if ("-javax-crypto-cipher".equalsIgnoreCase(arg))
                 benchmarkJavaxCryptoCipher = true;
-            else if ("-nio-block-cipher".equalsIgnoreCase(arg))
-                benchmarkNIOBlockCipher = true;
         }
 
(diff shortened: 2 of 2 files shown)


lyubomir:
> Broken Ant build

No:

1. Both the Ant and the Maven build files explicitly exclude CryptoBenchmark. I'm building libjitsi using Maven multiple times a day and it's not broken. I just tested the Ant build and it's not broken either.

2. https://github.com/jitsi/libjitsi/issues/132 talks about Maven (I asked there how the reporter reproduced the break), not Ant.

ibauersachs:
@lyubomir this file is in the main source tree, so IMO it should be compilable. Can we exclude the .class from the jar instead of excluding the .java?

In general, is the Ant script still needed? I'm only asking about the Java part, not the JNI and Debian scripts.

lyubomir:
@ibauersachs, I'm working on merging this PR. I merely corrected some of the incorrect statements in this PR and the respective issue.

pom.xml
```diff
@@ -83,11 +83,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
-          <excludes>
-            <exclude>org/jitsi/impl/neomedia/transform/srtp/CryptoBenchmark.java</exclude>
-          </excludes>
-        </configuration>
       </plugin>
```
lyubomir: I'm working on merging this PR so please don't touch it. But I want to state that I don't like it because it's incomplete and I have to spend non-trivial merge time on a non-issue (or at least an issue with very low priority). (1) Since you've removed the configuration tag, the whole plugin tag has become obsolete. (2) The discussion of whether the Ant build script is necessary is a different question out of the scope of this PR so, since you've removed the exclusion in the Maven file, it should be removed from the Ant file as well. Again, I'll do these so that I can merge this PR and be done with it.

PR #110, opened 2016-03-18
Enforce Perfect Forward Secrecy for DTLS

champtar:
As discussed when we were implementing DTLS cert cache (#98),
we want to enforce Perfect Forward Secrecy,
so that server compromise as no effect on past conferences.

On the client side Firefox is already enforcing it since v38
https://hacks.mozilla.org/2015/02/webrtc-requires-perfect-forward-secrecy-pfs-starting-in-firefox-38/
(don't know about chrome)

Tested with Firefox 44, Firefox Nightly 47 and Chrome 49

PR #134, opened 2016-04-10
Configurable DTLS certificate generation parameters.

bickelj:
This pull request is a replacement of https://github.com/jitsi/libjitsi/pull/67 which is now obsolete.

Here is a sample of property settings made available and being set:
org.jitsi.impl.neomedia.transform.dtls.RSA_KEY_SIZE_CERTAINTY=80
org.jitsi.impl.neomedia.transform.dtls.RSA_KEY_SIZE=2048
org.jitsi.impl.neomedia.transform.dtls.CERT_CACHE_EXPIRE_TIME=60000
<s>org.jitsi.impl.neomedia.transform.dtls.CERT_VALIDITY_TIME=43200000</s>

The jitsi-meet server administrator can set the number of key bits (...RSA_KEY_SIZE), set the certificate cache expire time (...CERT_CACHE_EXPIRE_TIME=0 for the "old" way should mean new certificate on every connection), and how long the certificate is valid for after the time of generation (...CERT_VALIDITY_TIME is above set to 12 hours). Times are in milliseconds.

Defaults remain the same as in master branch.

@ibauersachs and @champtar and @lyubomir please give your feedback on what can be improved in this pull request.

I have tested the above values using a properties file only and verified via wireshark to my server. I have not yet tried cache expiry of 0 yet, though I am pretty confident that will work too.

Thank you for considering this one, and you should already have a CLA on file.

One more note: had to add getLong to ConfigUtils in order to call it. That's why it's in here.

### DIFF
--- a/src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
+++ b/src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
@@ -78,12 +78,106 @@ public class DtlsControlImpl
     private static final long ONE_DAY = 1000L * 60L * 60L * 24L;
 
     /**
-     * The name of the property which specifies the signature algorithm used 
-     * during certificate creation. When a certificate is created and this 
+     * The name of the property which specifies the signature algorithm used
+     * during certificate creation. When a certificate is created and this
      * property is not set, a default value of "SHA1withRSA" will be used.
      */
-    public static final String PROP_SIGNATURE_ALGORITHM = 
+    public static final String PROP_SIGNATURE_ALGORITHM =
         "org.jitsi.impl.neomedia.transform.dtls.SIGNATURE_ALGORITHM";
+
+    /**
+     * The name of the property to specify RSA Key length.
+     */
+    public static final String RSA_KEY_SIZE_PNAME =
+        "org.jitsi.impl.neomedia.transform.dtls.RSA_KEY_SIZE";
+
+    /**
+     * The default RSA key size when configuration properties are not found.
+     */
+    public static final int DEFAULT_RSA_KEY_SIZE = 1024;
+
+    /**
+     * The RSA key size to use.
+     * The default value is {@code DEFAULT_RSA_KEY_SIZE} but may be overridden
+     * by the {@code ConfigurationService} and/or {@code System} property
+     * {@code RSA_KEY_SIZE_PNAME}.
+     */
+    public static final int RSA_KEY_SIZE;
+
+    /**
+     * The name of the property to specify RSA key size certainty.
+     * https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html
+     */
+    public static final String RSA_KEY_SIZE_CERTAINTY_PNAME =
+        "org.jitsi.impl.neomedia.transform.dtls.RSA_KEY_SIZE_CERTAINTY";
+
+    /**
+     * The RSA key size certainty to use.
+     * The default value is {@code DEFAULT_RSA_KEY_SIZE_CERTAINTY} but may be
+     * overridden by the {@code ConfigurationService} and/or {@code System}
+     * property {@code RSA_KEY_SIZE_CERTAINTY_PNAME}.
+     * For more on certainty, look at the three parameter constructor here:
+     * https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html
+     */
+    public static final int RSA_KEY_SIZE_CERTAINTY;
+
+    /**
+     * The default RSA key size certainty when config properties are not found.
+     */
+    public static final int DEFAULT_RSA_KEY_SIZE_CERTAINTY = 80;
+
+    /**
+     * The name of the property to specify DTLS certificate cache expiration.
+     */
+    public static final String CERT_CACHE_EXPIRE_TIME_PNAME =
+        "org.jitsi.impl.neomedia.transform.dtls.CERT_CACHE_EXPIRE_TIME";
+
+
+    /**
+     * The certificate cache expiration time to use, in milliseconds.
+     * The default value is {@code DEFAULT_CERT_CACHE_EXPIRE_TIME} but may be
+     * overridden by the {@code ConfigurationService} and/or {@code System}
+     * property {@code CERT_CACHE_EXPIRE_TIME_PNAME}.
+     */
+    public static final long CERT_CACHE_EXPIRE_TIME;
+
+    /**
+     * The default certificate cache expiration time, when config properties
+     * are not found.
+     */
+    public static final long DEFAULT_CERT_CACHE_EXPIRE_TIME = ONE_DAY;
+
+    /**
+     * The name of the property to specify DTLS certificate validity time.
+     */
+    public static final String CERT_VALIDITY_TIME_PNAME =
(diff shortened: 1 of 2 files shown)


bickelj:
Hi @jitsi-jenkins I have already signed the CLA, thanks.

src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
```diff
@@ -397,7 +522,7 @@ private static AsymmetricCipherKeyPair generateKeyPair()
         {
             long now = System.currentTimeMillis();
             Date notBefore = new Date(now - ONE_DAY);
-            Date notAfter = new Date(now + 6 * ONE_DAY);
+            Date notAfter = new Date(now + CERT_VALIDITY_TIME);
```
champtar: maybe we can use ```CERT_VALIDITY_TIME + ONE_DAY``` here

src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
```diff
@@ -518,14 +643,14 @@ public DtlsControlImpl(boolean srtpDisabled)
...
-        // we've decided to reuse their return values within a certain time
+        // intensive to invoke for each new DtlsControlImpl instance. That's
+        // why we've decided to reuse their return values within a certain time
         // frame. Attempt to retrieve from the cache.
         synchronized (DtlsControlImpl.class)
         {
             certificateInfo = certificateInfoCache;
             if (certificateInfo == null
-                    || certificateInfo.timestamp + ONE_DAY
+                    || certificateInfo.timestamp + CERT_CACHE_EXPIRE_TIME
```
champtar: and ```CERT_VALIDITY_TIME``` here, dropping ```CERT_CACHE_EXPIRE_TIME```, and changing comments ...

bickelj: In my view, the certificate cache expiration and the certificate expiration are loosely related but really are different concepts.
What if I want the certificate to expire after one day, but I want the cache to expire after 1 minute?

bickelj: The comment change was an attempt to conform to the style standards of no more than 80 characters per line.

src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
bickelj: I suppose for the same reason as (now - ONE_DAY)?
Which I am guessing is to avoid timezone or clock setting issues, so I can see this one.

But if someone bothers to change it, why not give them the ability to choose?

src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
champtar: in that case why aren't you adding yet another option for ```notBefore```?

bickelj: I'm open to that idea, it just wouldn't occur to me as something to change.

I see your point though, if using + ONE_DAY, then we can re-use the cache expiration for certificate expiration as before.

bickelj:
Is the purpose of re-using a single configuration setting to avoid allowing a conflicting pair of values that could cause trouble on the server? If so, could an assert statement with a friendly explanation of the difference solve the issue?

The purpose of disentangling the two settings would be to allow the server administrator to set the cache expiration without affecting the (already decent) default of certificate expiration.

My preference would be to be able to set the cache expiration to 0 if I want, or 2 days if I want. I see the point that if I set the cache to 100 days, then unless I also change the certificate expiration I will run into an issue.

Is that the idea?

src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
champtar: when i said 'changing comments' i meant if we use only one variable we need to adjust variable comments

bickelj: Oh, understood.

bickelj: What do you think of this: dropping CERT_VALIDITY_TIME and keeping CERT_CACHE_EXPIRE_TIME, and padding the validity by ONE_DAY to either side of the cache expiration?
Essentially the same idea, right? Except making clear to the admin that it is a cache that he/she is tuning. And then the software adjusts the certificate expirations to keep the admin out of trouble.

champtar:
I prefer simple config that always work, so having only one option help here.

Generating RSA key too often is useless as we now enforce PFS since #110.
On a webserver you regenerate your cert (including key) every month or less and nobody think it's insecure (for letsencrypt it's 90 days).
Chrome cache his cert for 30 days.

Maybe cert/cache validity should be in days and not in seconds, with a minimum of 1.

src/org/jitsi/impl/neomedia/transform/dtls/DtlsControlImpl.java
champtar: yep, or rename it CERT_CACHE_VALIDITY_TIME?

bickelj:
I hear what you're saying, and I cannot come up with any cybercryptosecuritygeek reason to disagree.

My priority is to chat with friends and family, so I don't really need to worry about my precious server processor being taxed, and I don't mind it being used for extraneous RSA key generation. I have no evidence, but I just have a personal emotional feeling that since the cost of generating new RSA keys in this instance is very low to me, I feel like generating them frequently.

If I were running a server with a large user-base I would be thinking more along the terms you describe. Or if I were a proprietary software vendor I might be worried if the people I was trying to enchant had to wait 6 seconds instead of 4 seconds.

But since I am just a friends-and-family user, and since I may be wanting to spend my own server power (that I pay for) to do RSA key generation, I think it's OK to offer me and other users that option, especially since it was the default for a long time and it worked fine.

bickelj:
Added https://github.com/jitsi/libjitsi/pull/134/commits/de02d9ad8e2aa430c91da87c5a12dd8c407bbed2 to the PR to simplify the configuration options.

I will say that it's not perfect because the user could put in a negative value, and perhaps the units are slightly cumbersome, but the point of this pull request is to allow the user to make inefficient or seemingly stupid choices if the user so desires. The defaults are left alone, those are a different discussion.

Choices now include three configuration options: specifying bit length of RSA keys, specifying the second argument to [BigInteger](https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#BigInteger%28int,%20int,%20java.util.Random%29) during key generation, and specifying how long the cache lives.

@ibauersachs thanks for your patience on this one, I hope you will consider it.

champtar:
Wait to hear from other, but i would squash the 2 commits into 1

bickelj:
Last time I attempted to squash I had a bad experience. But I'm probably just doing it wrong.

ibauersachs:
Jenkins: test

ibauersachs:
@bickelj The CLA message is automated, you can ignore it.

damencho:
Jenkins: it is ok to test

bickelj:
Thanks @champtar and @ibauersachs, appreciate you both.

PR #137, opened 2016-04-11
Ignore OpenSSL AES test when not running on Linux

ibauersachs:
The jnopenssl JNA library is only available for Linux. So don't attempt
to run the test anywhere else.


damencho:
Jenkins: it is ok to test

ibauersachs:
Jenkins: test this please

PR #139, opened 2016-04-14
SynchronizerImpl: use ConcurrentHashMap

champtar:
HashMap are not thread safe, even when reading,
so we can't use them outside of synchronized blocks,
else we can have some random crash or data/result
inconsistencies between threads.

ConcurrentHashMap is a well optimised beast, so use it.


champtar:
see #147

PR #141, opened 2016-04-14
Ensure localSourceID is positive

champtar:
Math.abs(Long.MIN_VALUE) == Long.MIN_VALUE which is negative
https://docs.oracle.com/javase/7/docs/api/java/lang/Math.html#abs%28long%29

Before this commit
 0 <= localSourceID < Integer.MAX_VALUE
 or localSourceID == (Long.MIN_VALUE % Integer.MAX_VALUE) == -2
After
 0 <= localSourceID <= Integer.MAX_VALUE

This was found using FindBugs
http://findbugs.sourceforge.net/bugDescriptions.html#RV_ABSOLUTE_VALUE_OF_RANDOM_INT

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -173,8 +173,7 @@ public class MediaStreamImpl
      * {@link this.rtpManager}, given that it offers this information with its
      * getLocalSSRC() method? TAG(cat4-local-ssrc-hurricane)
      */
-    private long localSourceID
-        = Math.abs(new Random().nextLong()) % Integer.MAX_VALUE;
+    private long localSourceID = (new Random().nextInt()) & 0x000000007FFFFFFFL;
 
     /**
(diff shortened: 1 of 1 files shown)


bgrozev:
We shouldn't be using Integer.MAX_VALUE here at all, since SSRCs are 32bit. What do you think about changing 0x7FFFFFFF to 0xFFFFFFFF?

champtar:
@bgrozev I agree we could use 0xFFFFFFFF, but this is taking an unnecessary risk (if somewhere we don't handle it well), and 0x7FFFFFFF is enough.

This commit was intended to prevent a close to impossible bug (and remove a line from FindBugs), not to possibly introduce new bugs for no gain :)

bgrozev:
Agreed, but I think that while we're at it we should also fix the other bug. But I can open a separate PR for that, if you like.

champtar:
Yep, i prefer a separate PR.

This is just an exemple, but if a piece of code is used in Jitsi, we should not break communication between old Jitsi and up to date Jitsi just to find some bug, ie deployment is slow and i prefer to keep back compat

PR #143, opened 2016-04-14
Use byte instead of int for the constant time comparison

champtar:
no need to expend byte into int each round


champtar:
forget about this, ```tempStore[i] ^ tagStore[i]``` gives an int, so keeping ```nonEqual``` as int is the way to go
http://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2

PR #145, opened 2016-04-14
Implements transactional key frame requests.

gpolitis:


src/org/jitsi/impl/neomedia/codec/video/Utils.java
```diff
@@ -36,8 +36,13 @@
     /**
      * Utility method that determines whether or not a packet is a key frame.
      */
-    public static boolean isKeyFrame(RawPacket pkt, Byte redPT, byte vp8PT)
+    public static boolean isKeyFrame(RawPacket pkt, Byte redPT, Byte vp8PT)
```
brianh5: if this function is vp8 specific, why not have it in the vp8 subdir?

emcho:
@brianh5 agreed. we should make sure that at least the part that assumes VP8 is properly abastracted.

emcho:
Just to be clear: I don't think we need to block this until we do, but we should add it on the next day todo lists.

src/org/jitsi/impl/neomedia/codec/video/Utils.java
gpolitis: Thanks @brianh5, you're right, this should be in the vp8 subdir.

src/org/jitsi/impl/neomedia/rtp/translator/RTPTranslatorImpl.java
```diff
@@ -225,6 +227,12 @@ public static int readUnsignedShort(byte[] buf, int off)
     private long localSSRC = -1;
 
     /**
+     *
+     */
+    private final RecurringProcessibleExecutor recurringProcessibleExecutor
```
lyubomir: @gpolitis, please add javadocs on the RecurringProcessibleExecutor field and getter. I know what RecurringProcessibleExecutor does in general but I'd like it to be obvious why RTPTranslatorImpl is a provider of RecurringProcessibleExecutor.

gpolitis: Will do, thank you @lyubomir !

src/org/jitsi/impl/neomedia/rtp/translator/RTPTranslatorImpl.java
```diff
@@ -585,6 +593,11 @@ int didRead(
                 int pt = buf[off + 1] & 0x7f;
 
                 format = streamRTPManager.getFormat(pt);
+
+                // Pass the packet to the feedback message sender to update
+                // its transactions.
+                rtcpFeedbackMessageSender.maybeStopRequesting(
+                    streamRTPManager, ssrc, pt, buf, off, len);
```
lyubomir: @gpolitis, "Pass the packet to the feedback message sender" is what the source code says so I find it stated in the comment unhelpful. I would like to read in the comment why it's necessary and, while "update its transactions" is in that spirit, I have no idea what transactions means. I guess I expect something more accessible. Anyway, that's my opinion and you don't have to act on it.

gpolitis: Will do, thank you @lyubomir !

src/org/jitsi/impl/neomedia/rtp/translator/RTCPFeedbackMessageSender.java
```diff
@@ -40,11 +72,17 @@
     private final RTPTranslatorImpl rtpTranslator;
 
     /**
-     * The next (command) sequence numbers to be used for RTCP feedback message
-     * packets per SSRC of packet sender-SSRC of media source pair.
+     * The {@link RecurringProcessibleExecutor} which will periodically call
+     * call {@link KeyframeRequester#process()} and trigger their retry logic.
```
lyubomir: will periodically call call

Could you please remove the extra occurrence of the word call, @gpolitis?

src/org/jitsi/impl/neomedia/rtp/translator/RTCPFeedbackMessageSender.java
```diff
@@ -102,87 +113,287 @@ private long getSenderSSRC()
...
-         * RED).
-         */
-        for (StreamRTPManager streamRTPManager
-                : rtpTranslator.getStreamRTPManagers())
-        {
-            MediaStream stream = streamRTPManager.getMediaStream();
+        // It's OK for this method to be a little "slow" (so that the
+        // {@code RTCPFeedbackMessageSender#maybeStopRequesting} is lock-less).
+        KeyframeRequester oldRequester = kfRequesters.putIfAbsent(
+            mediaSenderSSRC, new KeyframeRequester(mediaSenderSSRC));
```
lyubomir: I'm afraid I don't get it... You're going to initialize a new KeyframeRequester instance each time you want to send a FIR even if you don't really need that instance? I don't feel comfortable about that. Could you please explain whether that's the case?

gpolitis: I wanted to use a ConcurrentHashMap instead of a synchronized block (which would also have to be present in the `maybeStopRequesting` that runs for every incoming RTP packet). My thinking was that sending FIRs is much less often than RTP packets, so it's okay for the sendFIR method to be slow. I don't mind changing this, if you prefer it the other way.

src/org/jitsi/impl/neomedia/rtp/translator/RTCPFeedbackMessageSender.java
```diff
@@ -102,87 +113,287 @@ private long getSenderSSRC()
...
+         *
+         * @param mediaSenderSSRC
+         */
+        public KeyframeRequester(int mediaSenderSSRC)
         {
-            if (sendFIR(destination, mediaSenderSSRC))
-                sentFIR = true;
+            super(FIR_RETRY_INTERVAL_MS);
+            this.mediaSenderSSRC = mediaSenderSSRC;
+            this.sequenceNumber = new AtomicInteger(0);
```
lyubomir: I'm going to make a very moot point here, I admit. But isn't that easier to read if you initialize sequenceNumber above given that it's final and doesn't get initialized with a constructor argument?

src/org/jitsi/impl/neomedia/rtp/translator/RTCPFeedbackMessageSender.java
gpolitis: Will do, thank you @lyubomir !

src/org/jitsi/impl/neomedia/rtp/translator/RTCPFeedbackMessageSender.java
gpolitis: Will do, thank you @lyubomir !

src/org/jitsi/impl/neomedia/rtp/translator/RTCPFeedbackMessageSender.java
gpolitis: You're right, I now synchronize the writers to that map and keep the readers lock-less.

src/org/jitsi/impl/neomedia/rtp/translator/RTCPFeedbackMessageSender.java
```diff
@@ -102,87 +113,292 @@ private long getSenderSSRC()
...
+                }
+            }
+
+            if (vp8PT == null || vp8PT != pt)
+            {
+                return;
+            }
+
+            RawPacket pkt = new RawPacket(buf, off, len);
+            if (!Utils.isKeyFrame(pkt, redPT, vp8PT))
```
lyubomir: I get it that the JRE is optimized for small short-lived allocations. But I still don't feel absolutely at piece with allocating a new RawPacket just to pass it to a util and then leave it to the garbage collector... In my mind, having an overloaded Utils.isKeyFrame which takes byte[],int,int (in addition to the existing one which takes RawPacket) is OK. Anyway, I don't want you to do anything and I merely wanted to share what's on my mind.

gpolitis: You're right, it doesn't make any sense to allocate a `RawPacket` here. Thank you @lyubomir, I'll fix it!

src/org/jitsi/impl/neomedia/rtp/translator/RTCPFeedbackMessageSender.java
```diff
@@ -102,87 +113,291 @@ private long getSenderSSRC()
...
-        /*
-         * XXX Currently, the method effectively broadcasts a FIR i.e. sends it
-         * to all streams connected to rtpTranslator because MediaStream's
-         * method getRemoteSourceIds returns an empty list (possibly due to
-         * RED).
-         */
-        for (StreamRTPManager streamRTPManager
-                : rtpTranslator.getStreamRTPManagers())
+        boolean registerRecurringProcessible = false;
+        synchronized (kfRequesters)
```
bickelj: Could synchronization be removed by replacing this line and the following "if" with a call to atomic putIfAbsent(...)?

gpolitis: Hi @bickelj, yes, at the cost of having to create a new `KeyframeRequester` every time we attempt to send an FIR. But we only need a limited number of `KeyframeRequester`s per conference (because they're reused), so creating new ones over and over again seems like a waste of CPU cycles.

gavllew: The normal non-synchronised approach using ConcurrentHashMap would be:

    KeyframeRequester kfRequester = kfRequesters.get(mediaSenderSSRC);
    if (kfRequester == null)
    {
        // No existing requester; create a new one
        kfRequester = new KeyframeRequester(mediaSenderSSRC);
        KeyframeRequester existing = kfRequesters.putIfAbsent(mediaSenderSSRC, kfRequester);
        if (existing != null)
        {
            // Another thread created one at the same time, and beat this one
            kfRequester = existing;
        }
    }

This way you avoid locking, but you don't continue to create new instances long after one becomes available.

champtar: @gpolitis if you use 'get' you should never use 'containsKey', it's just 2 lookups instead of 1

gpolitis: Thank you all, you're right. @gavllew would you like to submit a PR for that?

gavllew: @gpolitis I'd love to, but so far trying to get agreement on the CLA from my employer's legal department has proved challenging.  I think my contribution above is small enough to avoid any sort of copyright dispute. :)

gpolitis: I understand, thank you nonetheless @gavllew for your valuable suggestion. @bickelj, @champtar would any of you be interested in providing a PR for this? Your suggestions make me think that you had something similar in your mind.

bickelj: @gpolitis I hadn't considered the creation of the new KeyframeRequester, was just trying to understand the reasoning, and your response makes sense. But now that @gavllew and @champtar have given further hints I can create a PR if no one else beats me to it.

PR #147, opened 2016-04-18
Synchronizer: concurrency fixes

champtar:
Except if you use Atomic* or volatile, you have to lock/synchronize for both writing and reading.


ibauersachs:
The first commit is fine. In the second one, the unsynchronized that were there before reads are okay because the conditions is checked against after obtaining a lock. The valid changes in the second commit are thus only the parameter validations.

champtar:
This type of check is only ok if you have an initial state (set in the constructor) and you never go back to it, and that's not the case here (see removeMapping).
synchronised ensure atomicity but also visibility between threads

ibauersachs:
@bgrozev informed me that this code isn't even in use. As I'm lacking time to go through it again, I'm closing this PR.

champtar:
@ibauersachs this was spotted with FindBugs and haven't checked that it's used

PR #148, opened 2016-04-19
Less locking but also less unnecessary new objects.

bickelj:
In response to @gpolitis and conversation [here](https://github.com/jitsi/libjitsi/pull/145/files/7cdebacba6539fab0a9ae8a2f715889223009ce9#r60075218)

I didn't see a quick and easy way to test this, but basic sanity checks seem to pass (e.g. the videobridge runs and serves video).


champtar:
To keep modifications to a minimum you could have used ```kfRequester``` instead of ```keyframeRequester```, except that it looks ok for me.

Side note: this is ok because ```new KeyframeRequester(mediaSenderSSRC)``` has no side effect

PR #150, opened 2016-04-19
Moves AudioSilence{Media,Capture}Device to libjitsi.

bgrozev:
Moves the "audio silence" devices from jitsi-videobridge to libjitsi,
because they are also used in jitsi-hammer.


bgrozev:
Jenkins: ok to test

PR #152, opened 2016-04-25
Rework getRoundTripDelay to reduce 'RTT computation may be wrong'

champtar:
If RTT is really small (<1ms, like in a test setup), we might
end up with negative ntprtd value, but less than 1ms, which
we can round to 0. This is normal because timeMs is System.currentTimeMillis(),
so the (im)precision is 1ms at best.


bgrozev:
This has been bugging us for a while. Thank you, @champtar!

PR #153, opened 2016-04-26
Not logging a packet should be a warning.

gpolitis:


ibauersachs:
This could lead to flooded logfiles and the messages aren't particularly helpful.

gpolitis:
This transform engine is not enabled by default as it's mainly used for debugging. When it is enabled, you want to know whether or not a packet has been successfully logged or not.

PR #155, opened 2016-05-05
Add DelayingTransformEngine

paweldomas:


champtar:
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/DelayQueue.html
Just asking why a number of packet and not a time?

paweldomas:
This is going to be used only for audio/video synchronisation tests(lip sync) where we will be delaying audio packets and then trying to see if the video has been delayed as well. We do not do any calculations based on this packet delay value. This way it's just easier and simpler.

bgrozev:
@champtar I agree that using a DelayQueue is a better approach in general, but this solution is simpler and works for our use-case.

PR #167, opened 2016-06-07
Tries to fix the ptime attribute for opus.

bgrozev:


ibauersachs:
Does that fix jitsi/jitsi#266?

ibauersachs:
I'm not an SDP specialist, but if the old output was:
```
a=fmtp:96 usedtx=1
a=ptime:96 20
```
then with this patch, the new output would be `a=fmtp:96 usedtx=1; ptime=20`. But RFC7587 says they must be like this if present:
```
a=fmtp:96 usedtx=1
a=ptime:20
a=maxptime:40
```

Am I missing something?

bgrozev:
You are right, Ingo. I didn't look at the spec, and naively thought the format is the one that is obvious and makes sense.

ibauersachs:
RFC != makes sense :(

PR #171, opened 2016-06-22
libjitsi changes for sequence number rewriting for video packets.

mdaneshi:
This is PR includes the libjitsi changes required for sequence number rewriting. In addition to some helper functions, we a seq number reset logic is added to make sure the rewriting works fine.

### DIFF
--- a/src/org/jitsi/impl/neomedia/RawPacket.java
+++ b/src/org/jitsi/impl/neomedia/RawPacket.java
@@ -865,6 +865,20 @@ public class RawPacket
     }
 
     /**
+     * Set sequence number for an RTP buffer
+     *
+     * @param buffer
+     * @param offset
+     * @param seq
+     *
+     */
+    public static void setSequenceNumber(byte[] buffer, int offset, int seq)
+    {
+        buffer[offset + 2] = (byte) (seq>>8 & 0xff);
+        buffer[offset + 3] = (byte) (seq & 0xff);
+    }
+
+    /**
      * Get SRTCP sequence number from a SRTCP packet
      *
      * @param authTagLen authentication tag length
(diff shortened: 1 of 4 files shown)


brianh5:
i think maryam should already be covered by our CLA.

bgrozev:
Yes, we have the CLA.

Jenkins: add to whitelist

src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java
```diff
@@ -205,6 +207,8 @@ private int doWrite(
             OutputDataStreamDesc s = streams.get(i);
             StreamRTPManagerDesc streamRTPManager
                 = s.connectorDesc.streamRTPManagerDesc;
+            // Reset the sequence number to the original value
```
bgrozev: Can you please describe in more detail what's going on here (that there's a contract that writing may result in a change the sequence number but not anything else, and what the reasons are). I really think that's a tricky thing and worth documenting.

src/org/jitsi/impl/neomedia/RawPacket.java
```diff
@@ -865,6 +865,20 @@ public static int getSequenceNumber(byte[] buffer, int offset, int length)
...
 
     /**
+     * Set sequence number for an RTP buffer
+     *
+     * @param buffer
+     * @param offset
+     * @param seq
+     *
+     */
+    public static void setSequenceNumber(byte[] buffer, int offset, int seq)
```
bgrozev: Can you please also change setSequenceNumber(int) so that it uses this method?

src/org/jitsi/util/RTPUtils.java
```diff
@@ -36,4 +36,9 @@ else if (diff > 1<<15)
 
         return diff;
     }
+
```
bgrozev: Can you please add javadocs? It is required by our code convention.

PR #173, opened 2016-06-27
Use colons instead of dashes in a=imageattr lines.

josephfrazier:
Also, starts ranges with 1 instead of 0.

Links to RFC 6236 instead of draft-ietf-mmusic-image-attributes-04.

This fixes the Jitsi bug reported here:
http://lists.jitsi.org/pipermail/users/2016-June/011253.html

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaUtils.java
+++ b/src/org/jitsi/impl/neomedia/MediaUtils.java
@@ -610,7 +610,7 @@ public class MediaUtils
     /**
      * Creates value of an imgattr.
      *
-     * http://tools.ietf.org/html/draft-ietf-mmusic-image-attributes-04
+     * https://tools.ietf.org/html/rfc6236
      *
      * @param sendSize maximum size peer can send
      * @param maxRecvSize maximum size peer can display
@@ -632,10 +632,10 @@ public class MediaUtils
             img.append(",y=");
             img.append((int)sendSize.getHeight());
             img.append("]");*/
-            /* send [x=[min-max],y=[min-max]] */
-            img.append("send [x=[0-");
+            /* send [x=[min:max],y=[min:max]] */
+            img.append("send [x=[1:");
             img.append((int)sendSize.getWidth());
-            img.append("],y=[0-");
+            img.append("],y=[1:");
             img.append((int)sendSize.getHeight());
             img.append("]]");
             /*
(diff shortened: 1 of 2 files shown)


ibauersachs:
Do you see any way so that this doesn't break compatibility with current/old Jitsi clients? (It would already be the case if we had implemented the entired draf-04, but, well... :()

josephfrazier:
Yeah... I'm not sure if there's a way to maintain backwards compatibility, unfortunately :/

Do you know if the current/old clients would simply ignore the lines that have the new syntax? If so, is that an acceptable level of "compatibility"? If not, does Jitsi have "major" releases that are allowed to be backwards incompatible, where a change like this could land?

ibauersachs:
I don't know what currently happens with an invalid line: parseSendRecvResolution always creates an array and the usages of this method only check if the result is null (which is never the case) :-(

There aren't really any policies of breaks with previous stables. Can you please update the parsing code so that newer versions at least can handle older client (i.e. understand the `-` syntax)? Should be as simple as changing the regex to `...(-|:)...`.

Please also take note of the CLA message above. Sorry if you've already signed it, but I don't have access to the list of signees.

josephfrazier:
Thanks for the feedback, I changed the parsing regex as suggested. However, I haven't yet been able to test compatibility between old/new Jitsi clients, because I cannot run multiple Jitsi instances on the same machine. 

I signed the CLA shortly before submitting the PR. Perhaps someone else can verify?

ibauersachs:
@damencho thanks:

@josephfrazier you can set an alternate home directory for Jitsi to launch a second instance with `-Dnet.java.sip.communicator.SC_HOME_DIR_NAME=Jitsi-dev` (assuming you're launching it from Eclipse, add it to the VM arguments as described in https://jitsi.org/Documentation/ConfigureEclipseNew).

Please ket me know if you still want to test, otherwise I'm just going to merge it and we'll see what happens.

josephfrazier:
Thanks for the testing instructions. I'm using ant instead of Eclipse, but I couldn't figure out how to pass the argument properly, so I just edited build.xml as follows before `ant run`:

```diff
diff --git a/build.xml b/build.xml
index 1be362d..6a57020 100644
--- a/build.xml
+++ b/build.xml
@@ -849,8 +849,9 @@

             <!--sysproperty key="net.java.sip.communicator.SC_HOME_DIR_LOCATION"
                 value="${user.home}"/>
+                -->
             <sysproperty key="net.java.sip.communicator.SC_HOME_DIR_NAME"
-                value="schome"/-->
+                value="Jitsi-dev"/>

             <sysproperty key="smack.debugEnabled"
                 value="${smack.debugEnabled}"/>
```

I tested H264 calls in both directions between old/new clients, and they appear to still work (implying that the old client ignores the new syntax).

ibauersachs:
Merged. Thank you!

josephfrazier:
Cheers! Is there anything I can do to help get these changes into a Jitsi release?

ibauersachs:
No, libjitsi needs to be updated as binary there. I think @bgrozev has a pending update anyway.

PR #175, opened 2016-07-06
Refactors Logger. Adds a second Logger implementation with more flexible

bgrozev:
level configuration.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>ice4j</artifactId>
+      <version>1.1-20160706.192041-13</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java
+++ b/src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java
@@ -29,6 +29,7 @@ import org.jitsi.service.configuration.*;
 import org.jitsi.service.libjitsi.*;
 import org.jitsi.service.packetlogging.*;
 import org.jitsi.util.*;
+import org.jitsi.util.Logger; // Disambiguation.
 
 /**
  *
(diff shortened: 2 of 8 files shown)


ibauersachs:
Please hold off with merging this, I'm a bit concerned about the package org.jitsi.util and making Logger abstract.

bgrozev:
Jenkins: test this please

bgrozev:
@ibauersachs , I'll hold.

bgrozev:
@ibauersachs FYI: I tested Jitsi after replacing the ice4j and libjitsi jars, and didn't notice any problems.

ibauersachs:
Do we know of any other libjitsi users that could be affected (i.e. outside of jitsi-x)?

And then there's the issue about org.jitsi.util... There are already multiple libraries (can't remember if there's more than the lgpl-deps) declaring this package and thus breaking OSGi. Moving more classes here doesn't help unless libjitsi is declared the master library for org.jitsi.util and the others get migrated at some point.

Otherwise: LGTM

bgrozev:
> Do we know of any other libjitsi users that could be affected (i.e. outside of jitsi-x)?

Unless I am missing something, this shouldn't affect users: the Logger constructor used to be private, and the Logger.getLogger(Class<?>) and Logger.getLogger(String) APIs are unchanged.

The ice4j change may break compilation because its new Logger class clashes with java.util.logging.Logger, but that's an easy fix.

> And then there's the issue about org.jitsi.util... There are already multiple libraries (can't remember if there's more than the lgpl-deps) declaring this package and thus breaking OSGi. Moving more classes here doesn't help unless libjitsi is declared the master library for org.jitsi.util and the others get migrated at some point.

I couldn't find org.jitsi.util in other places apart from lgpl-deps. As far as I know, libjitsi is the right place for these classes (there was some duplication with net.sip.communicator.util, but that's a separate package). If we need to add another util package in another repository, we should then make sure to use another package name (like org.jitsi.jicofo.util does).

ibauersachs:
> Unless I am missing something, this shouldn't affect users: the Logger constructor used to be private, and the Logger.getLogger(Class<?>) and Logger.getLogger(String) APIs are unchanged.

Right, I missed the private constructor.

> The ice4j change may break compilation because its new Logger class clashes with java.util.logging.Logger, but that's an easy fix.

The wildcard imports are a (bad) Jitsi practice, discouraged, and rarely used by other projects. Anyway, this is a compile-time only issue.

damencho:
Jenkins: test this please

damencho:
Jenkins: test this please

PR #176, opened 2016-07-06
RTP timestamp uplifting fixes

gpolitis:
Makes sure we update the minTimestamp value when we perform RTP
timestamp uplifting. Failure in doing so may result in maxTimestamp
not being updated, breaking RTP timestamp uplifting.


bgrozev:
The test failure is unrelated.

PR #179, opened 2016-07-19
Refactors seqnum rewriting for video packets.

gpolitis:
- Extracts the rewriting logic in a re-usable module.
- Reuses the module for rewriting sequence numbers of audio packets.
- Moves all the logic inside the OutputDataStreamImpl (as opposed to
  having half of it in the OutputDataStreamImpl and half of it in
  the VideoChannel).
- Adds tests for the seqnum rewriting.

### DIFF
--- a/src/org/jitsi/impl/neomedia/SequenceNumberRewriter.java
+++ b/src/org/jitsi/impl/neomedia/SequenceNumberRewriter.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright @ 2015 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.impl.neomedia;
+
+import org.jitsi.util.*;
+
+/**
+ * Rewrites sequence numbers for RTP streams by hiding any gaps caused by
+ * dropped packets. Rewriters are not thread-safe. If multiple threads access a
+ * rewriter concurrently, it must be synchronized externally.
+ *
+ * @author Maryam Daneshi
+ * @author George Politis
+ */
+public class SequenceNumberRewriter
+{
+    /**
+     * The delta between what's been accepted and what's been received, mod 16.
+     */
+    int delta = 0;
+
+    /**
+     * The highest sequence number that got accepted, mod 16.
+     */
+    int highestSent = -1;
+
+    /**
+     * Rewrites the sequence number of the RTP packet in the byte buffer, hiding
+     * any gaps caused by drops.
+     *
+     * @param accept true if the packet is accepted, false otherwise
+     * @param buf the byte buffer that contains the RTP packet
+     * @param off the offset in the byte buffer where the RTP packet starts
+     * @param len the length of the RTP packet in the byte buffer
+     */
+    public void rewrite(boolean accept, byte[] buf, int off, int len)
+    {
+        if (buf == null || buf.length + off < len)
+        {
+            return;
+        }
+
+        int sequenceNumber = RawPacket.getSequenceNumber(buf, off, len);
+        int newSequenceNumber = rewriteSequenceNumber(accept, sequenceNumber);
+
+        if (sequenceNumber != newSequenceNumber)
+        {
+            RawPacket.setSequenceNumber(buf, off, newSequenceNumber);
+        }
+    }
+
+    /**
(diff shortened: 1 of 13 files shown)


src/org/jitsi/impl/neomedia/transform/csrc/SsrcTransformEngine.java
```diff
@@ -261,6 +266,17 @@ public RawPacket reverseTransform(RawPacket pkt)
                 csrcAudioLevelDispatcher.addLevels(levels, pkt.getTimestamp());
             }
         }
+
+        SequenceNumberRewriter rewriter = ssrcToRewriter.get(pkt.getSSRCAsLong());
```
mdaneshi: @gpolitis The fact that the rewriter is accessed after the lastN decision means that not all packets will reach here (if write is "false" we never get here) Rewriter needs to see all the packets of all ssrcs.

bgrozev: Not sure I understand this. This is in the audio chain, so packets wouldn't have been dropped elsewhere, so all packets will reach the rewriter. Am I missing something?

bgrozev: Was the comment meant for the changes to OutputDataStreamImpl.java?

src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java
```diff
@@ -65,6 +65,12 @@
     private final RTPConnectorImpl connector;
 
     /**
+     * A map of source ssrc to {@link SequenceNumberRewriter}.
+     */
+    private final Map<Long, SequenceNumberRewriter> ssrcToRewriter
+        = new HashMap<>();
```
bgrozev: Don't we need a separate rewriter for each receiver?

src/org/jitsi/impl/neomedia/transform/csrc/SsrcTransformEngine.java
gpolitis: Thank you both for the comments! I'll update the code accordingly.

src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java
gpolitis: Right, part of the code needs to be in the `OutputStreamImplDesc`.

src/org/jitsi/impl/neomedia/SequenceNumberRewriter.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+        else
+        {
+            // update the delta (if needed)
+            if (highestSent != -1)
+            {
+                final int d
+                    = RTPUtils.subtractNumber(sequenceNumber, highestSent);
+
+                if (delta == 0 ||
```
bgrozev: Why special handling for delta==0? If delta == 0, d=2^16-1, we want delta to stay 0. Also, can we rename "d"?

gpolitis: Right 👍

gpolitis:
jenkins it's ok to test

gpolitis:
jenkins it's ok to test

src/org/jitsi/impl/neomedia/transform/csrc/SsrcTransformEngine.java
```diff
@@ -261,6 +267,17 @@ public RawPacket reverseTransform(RawPacket pkt)
...
+        if (rewriter == null)
+        {
+            rewriter = new SequenceNumberRewriter();
+            ssrcToRewriter.put(pkt.getSSRCAsLong(), rewriter);
+        }
+
+        rewriter.rewrite(
+            !dropPkt, pkt.getBuffer(), pkt.getOffset(), pkt.getLength());
```
bgrozev: This transformer is placed in the chain before SRTP. If it rewrites sequence numbers SRTP authentication will fail :(

src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java
```diff
@@ -250,17 +250,34 @@ private int doWrite(
...
                         /* source */ exclusion,
-                        buf, off, len,
+                    buf, off, len,
                         /* destination */ streamRTPManager,
-                        _data);
+                    _data);
+            }
+
+            // Hide gaps in the sequence numbers because of dropping packets.
+            Long ssrc = RawPacket.getSSRCAsLong(buf, off, len);
```
bgrozev: This code executes for RTCP as well. I suspect that up until now it somehow worked without major issues, but with this it will quickly fill the map since getSSRCAsLong will actually read the NTP timestamp if the packet is RTCP, and that changes with each packet.

bgrozev:
I think this is good to go now. @gpolitis can you please check if my changes make sense?

@brianh5 @mdaneshi Would you be able to test this in your scenario?

gpolitis:
LGTM

gpolitis:
Merging as this has been manually tested, the automated tests have passed and we need this deployed.

PR #181, opened 2016-07-27
RTP stream pause/resume improvements.

gpolitis:
Rewrites timestamps in addition to sequence numbers when a stream is paused/resumed.

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
@@ -37,7 +37,7 @@ public class RTCPHeaderUtils
      */
     public static int getPacketType(byte[] buf, int off, int len)
     {
-        if (buf == null || Math.min(buf.length, len) < off + 2)
+        if (buf == null || buf.length < off + Math.max(len, 2))
         {
             return -1;
         }
@@ -56,7 +56,7 @@ public class RTCPHeaderUtils
      */
     public static long getSenderSSRC(byte[] buf, int off, int len)
     {
-        if (buf == null ||  Math.min(buf.length, len) < off + 8)
+        if (buf == null || buf.length < off + Math.max(len, 8))
         {
             return -1;
         }
@@ -78,7 +78,7 @@ public class RTCPHeaderUtils
      */
     public static int getLength(byte[] buf, int off, int len)
     {
-        if (buf == null ||  Math.min(buf.length, len) < off + 4)
+        if (buf == null || buf.length < off + Math.max(len, 4))
         {
             return -1;
         }
@@ -100,7 +100,7 @@ public class RTCPHeaderUtils
      */
     public static int getVersion(byte[] buf, int off, int len)
     {
-        if (buf == null ||  Math.min(buf.length, len) < off + 1)
+        if (buf == null || buf.length < off + Math.max(len, 1))
         {
             return -1;
         }
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright @ 2015 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.impl.neomedia.rtcp;
+
+import net.sf.fmj.media.rtp.*;
+
+/**
+ * Utility class that contains static methods for RTCP header manipulation.
+ *
+ * TODO maybe merge into the RTCPHeader class.
+ *
+ * @author George Politis
+ */
+public class RTCPHeaderUtils
+{
+    /**
+     * Gets the RTCP packet type.
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return the unsigned RTCP packet type, or -1 in case of an error.
+     */
+    public static int getPacketType(byte[] buf, int off, int len)
+    {
+        if (buf == null || Math.min(buf.length, len) < off + 2)
+        {
+            return -1;
+        }
+
+        return buf[off + 1] & 0xff;
+    }
+
+    /**
+     * Gets the RTCP sender SSRC.
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return the unsigned RTCP sender SSRC, or -1 in case of an error.
+     */
+    public static long getSenderSSRC(byte[] buf, int off, int len)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 8)
+        {
+            return -1;
+        }
+
+        return (((buf[off + 4] & 0xff) << 24)
+            | ((buf[off + 5] & 0xff) << 16)
+            | ((buf[off + 6] & 0xff) << 8)
+            | (buf[off + 7] & 0xff)) & 0xffffffffl;
+    }
+
+    /**
+     * Gets the RTCP packet length in bytes.
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return  the RTCP packet length in bytes, or -1 in case of an error.
+     */
+    public static int getLength(byte[] buf, int off, int len)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 4)
+        {
+            return -1;
+        }
+
+        int lengthInWords
+            = ((buf[off + 2] & 0xff) << 8) | (buf[off + 3] & 0xff);
+
+        return (lengthInWords + 1) * 4;
+    }
+
+    /**
+     * Gets the RTCP packet version.
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return the RTCP packet version, or -1 in case of an error.
+     */
+    public static int getVersion(byte[] buf, int off, int len)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 1)
+        {
+            return -1;
+        }
+
+        return (buf[off] & 0xc0) >>> 6;
+    }
+
+    /**
+     * Checks whether the RTCP header is valid or not. It does so by checking
+     * the RTCP header version and makes sure the buffer is at least 8 bytes
+     * long.
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return true if the RTCP packet is valid, false otherwise.
+     */
+    public static boolean isValid(byte[] buf, int off, int len)
+    {
+        int version = RTCPHeaderUtils.getVersion(buf, off, len);
+        if (version != RTCPHeader.VERSION)
+        {
+            return false;
+        }
+
+        int pktLen = RTCPHeaderUtils.getLength(buf, off, len);
+        if (pktLen < RTCPHeader.SIZE)
+        {
+            return false;
+        }
+
+        return true;
+    }
+}
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
@@ -16,6 +16,7 @@
 package org.jitsi.impl.neomedia.rtcp;
 
 import net.sf.fmj.media.rtp.*;
+import org.jitsi.impl.neomedia.*;
 
 /**
  * Utility class that contains static methods for RTCP header manipulation.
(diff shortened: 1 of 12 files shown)


brianh5:
is this in response to https://bugs.chromium.org/p/webrtc/issues/detail?id=5868&can=1&q=is%3Astarred%20&colspec=ID%20Pri%20Mstone%20ReleaseBlock%20Component%20Status%20Owner%20Summary ? or simulcast streams stopping/starting. (or i suppose both).

gpolitis:
@brianh5 neither, it's much simpler than that :) When hiding the sequence number gaps (because of last N, for example), we also need to hide the gaps in the timestamps. Otherwise the jitter buffer might freak out. Does that make sense?

brianh5:
gotcha.

gpolitis:
jenkins it's ok to test.

src/org/jitsi/impl/neomedia/rtp/StreamRTPManager.java
```diff
@@ -56,10 +56,48 @@
     /**
      * A map of source ssrc to {@link ResumableStreamRewriter}.
      */
-    public final Map<Long, ResumableStreamRewriter> ssrcToRewriter
+    private final Map<Long, ResumableStreamRewriter> ssrcToRewriter
         = new HashMap<>();
 
     /**
+     *
```
bgrozev: Non-blocking: can we add documentation and move these after the constructor?

src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java
```diff
@@ -761,4 +740,80 @@ boolean writeControlPayload(Payload controlPayload, MediaStream destination)
...
+        {
+            int offset = off, length = len;
+
+            boolean modified = false;
+
+            while (length > 0)
+            {
+                // Check RTCP packet validity. This makes sure that pktLen > 0
+                // so this loop will eventually terminate.
+                if (RTCPHeaderUtils.isValid(buf, offset, length))
```
bgrozev: Shouldn't this break if the packet is not valid?

src/org/jitsi/impl/neomedia/rtp/ResumableStreamRewriter.java
```diff
@@ -135,6 +139,73 @@ public boolean rewrite(boolean accept, byte[] buf, int off, int len)
...
+        }
+
+        long ts = RTCPSenderInfoUtils.getTimestamp(buf, off, len);
+        if (ts == -1)
+        {
+            return false;
+        }
+
+        long newTs = rewrite
+            ? (ts - timestampDelta) & 0xffffffff
```
bgrozev: The constant should be a long (0xffffffffL)

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return the unsigned RTCP packet type, or -1 in case of an error.
+     */
+    public static int getPacketType(byte[] buf, int off, int len)
+    {
+        if (buf == null || Math.min(buf.length, len) < off + 2)
```
bgrozev: off needs to be included in the check for buf.length, but not for len. E.g. the following is OK:
buf.length=10000, off=5000, len=10

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return the unsigned RTCP sender SSRC, or -1 in case of an error.
+     */
+    public static long getSenderSSRC(byte[] buf, int off, int len)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 8)
```
bgrozev: Same as above

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return  the RTCP packet length in bytes, or -1 in case of an error.
+     */
+    public static int getLength(byte[] buf, int off, int len)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 4)
```
bgrozev: Same as above

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP header starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return the RTCP packet version, or -1 in case of an error.
+     */
+    public static int getVersion(byte[] buf, int off, int len)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 1)
```
bgrozev: Same as above

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
```diff
@@ -0,0 +1,138 @@
...
+    public static boolean isValid(byte[] buf, int off, int len)
+    {
+        int version = RTCPHeaderUtils.getVersion(buf, off, len);
+        if (version != RTCPHeader.VERSION)
+        {
+            return false;
+        }
+
+        int pktLen = RTCPHeaderUtils.getLength(buf, off, len);
+        if (pktLen < RTCPHeader.SIZE)
```
bgrozev: Should probably take the upper bound into account as well (len and buf.length)

src/org/jitsi/impl/neomedia/rtcp/RTCPSenderInfoUtils.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param buf the byte buffer that contains the RTCP header.
+     * @param off the offset in the byte buffer where the RTCP sender info
+     * starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @return the RTP timestamp, or -1 in case of an error.
+     */
+    public static long getTimestamp(byte[] buf, int off, int len)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 12)
```
bgrozev: Same as above

src/org/jitsi/impl/neomedia/rtcp/RTCPSenderInfoUtils.java
```diff
@@ -0,0 +1,99 @@
...
+        {
+            return -1;
+        }
+
+        return (((buf[off + 8] & 0xff) << 24)
```
bgrozev: Can we abstract this? Maybe use RawPacket.readInt()?

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
```diff
@@ -0,0 +1,138 @@
...
+     * @return the unsigned RTCP sender SSRC, or -1 in case of an error.
+     */
+    public static long getSenderSSRC(byte[] buf, int off, int len)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 8)
+        {
+            return -1;
+        }
+
+        return (((buf[off + 4] & 0xff) << 24)
```
bgrozev: Can we abstract this? Maybe use RawPacket.readInt()?

src/org/jitsi/impl/neomedia/rtcp/RTCPSenderInfoUtils.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     * @param ts the new timestamp to be set.
+     *
+     * @return the number of bytes written.
+     */
+    public static int setTimestamp(
+        byte[] buf, int off, int len, long ts)
+    {
+        if (buf == null ||  Math.min(buf.length, len) < off + 12)
```
bgrozev: Same as above

src/org/jitsi/impl/neomedia/rtcp/RTCPSenderInfoUtils.java
```diff
@@ -0,0 +1,99 @@
...
+        {
+            return -1;
+        }
+
+        buf[off + 8] = (byte)(ts>>24);
+        buf[off + 9] = (byte)(ts>>16);
+        buf[off + 10] = (byte)(ts>>8);
+        buf[off + 11] = (byte)ts;
+
+        return 12;
```
bgrozev: In what sense have we written 12 bytes? Given the current usage of this, it would be better to declare this void and return 'true' in processRTCP()

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
gpolitis: I think this is being taken cared of by `RTCPHeaderUtils.getVersion()` and `RTCPHeaderUtils.getLength()`.

src/org/jitsi/impl/neomedia/rtcp/RTCPSenderInfoUtils.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param off the offset in the byte buffer where the RTCP sender info
+     * starts.
+     * @param len the number of bytes in buffer which constitute the actual
+     * data.
+     *
+     * @return true if the RTCP sender info is valid, false otherwise.
+     */
+    public static boolean isValid(byte[] buf, int off, int len)
+    {
+        if (buf == null || Math.min(buf.length, len) < RTCPSenderInfo.SIZE)
```
bgrozev: Same as above

src/org/jitsi/impl/neomedia/transform/DiscardTransformEngine.java
```diff
@@ -42,40 +43,101 @@
...
-        return dropPkt ? null : pkt;
-    }
+            byte[] buf = pkt.getBuffer();
+            int offset = pkt.getOffset(), length =  pkt.getLength();
+
+            while (length > 0)
+            {
+                // Check RTCP packet validity. This makes sure that pktLen > 0
+                // so this loop will eventually terminate.
+                if (RTCPHeaderUtils.isValid(buf, offset, length))
```
bgrozev: Break if the packet is invalid?

src/org/jitsi/util/RTPUtils.java
```diff
@@ -38,6 +38,22 @@ else if (diff > 1<<15)
...
 
     /**
+     * Returns the difference between two RTP timestamps (modulo 2^32).
+     * @return the difference between two RTP timestamps (modulo 2^32).
+     */
+    public static long timestampDiff(long a, long b)
+    {
+        long diff = a - b;
+
+        if (diff < -(1<<31))
```
bgrozev: Any reason not to use TimeUtils.rtpDiff? If this needs to stay here, the "1" numerals need to be long to prevent overflows.

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
bgrozev: getVersion checks whether the length is at least 1 byte; getLength checks whether the length is at least 4 bytes. There is no check that what getLength returns is:
a) <= len
b) <= buf.length - off

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
gpolitis: (facepalm)

src/org/jitsi/impl/neomedia/rtcp/RTCPHeaderUtils.java
gpolitis: Now the buffer over/underflow checks are fixed, this should be ok.

brianh5:
@gpolitis looks like this may fix https://bugs.chromium.org/p/webrtc/issues/detail?id=6026 for us, which would be great.  are you doing anything to handle this scenario?

stream 1 stops being forwarded, last packets were:
{seq 1, ts: 10}, {seq 2, ts: 10}, {seq 3, ts: 10} [incomplete frame for ts 10]
...now a period where stream 1 isn't being forwarded...
...stream 1 starts being forwarded again, sequence and timestamp are obviously much further ahead but are re-written to match

in order to address the issue above (for h264, at least) we'd need to make the jitter buffer happy so it flushes correctly.  i think there are a couple options:
1) send the resumed packets with the same timestamp as the old one...this gets weird though, as with h264 it could throw things off with regards to the market bit, since the jitter buffer will be looking for the next packet with one to know when the frame ends.
2) send the resumed packets with the next anticipated timestamp: this won't work as it'll cause the same slow switching problem we observed when filing the bug.
3) remember if the last packet sent was the end of a frame or not...if it was, can send the new data with the next timestamp without tweaking anything.  if it wasn't, then send a dummy packet with the same timestamp as the last one to mark the end of the last frame?  sounds a bit hacky, but not sure how else to make the jitter buffer happy here.

gpolitis:
@brianh5 3) should work I guess, but it's yet another hack. Wouldn't a proper solution be to make sure that we stop on frame boundaries and resume with a keyframe? We already have some code for frame boundaries and key frame detection in the simulcast code that we can factor out.

brianh5:
@gpolitis i think that should work if the timestamps incremented correctly.  does seem a bit awkward to have to fulfill the rest of the frame even if that stream isn't technically 'interesting' to the client anymore (i.e. no longer in last-n), and may be awkward to have to either a) forward a stream that isn't in last-n to complete the frame or b) have the last-n temporarily go bigger than it should.

gpolitis:
You're right, it's awkward and it could also lead to bugs. Do you want to give 3) a try?

brianh5:
hey @gpolitis we're gonna think on it a bit...agree it's not a pretty solution.  also discussing on the bug i linked if there's a way that situation can be handled better on the jitter buffer side (though, admittedly, the timestamp rewriting may actually make that more difficult to deal with on the jitter buffer side since there won't be any timestamp gap to pick up on anymore).

bgrozev:
I'm not following closely, but I had the following idea and George doesn't find it absolutely ridiculous so I'm sharing: skip a sequence number when resuming (unless it stopped on a marked packet). The jb will not output the old frame since it knows it's incomplete, and as long as the next frame is decodable it should be output. It will trigger some NACKs though.

brianh5:
@bgrozev  i think the problem there is that the jb will wait for quite a while before giving up on the incomplete frame, which is what causes the slow switch.

PR #184, opened 2016-08-04
Auto keyframe control for recording (fast seek)

zbettenbuk:
This change enables video bridge users (or other custom libjitsi projects) that solely use the stack for server side recording to enable auto keyframe control for faster seeking in recorded files and automatically correct intermittent issues. The function is controlled by a system parameter.


zbettenbuk:
Agreement already signed (btw the links are broken in the comment)

damencho:
Yep, we know about the links, need update but it will happen next week.
Jenkins: add to whitelist

PR #190, opened 2016-08-16
Statistics improvements

bgrozev:
Refactors MediaStreamStats. Fixes the loss statistics and implements additional stream statistics.

Note that because the split into commits was done after the code was written the code may not even compile at each commit. The effort to avoid this is, IMO, not worth the benefits.


src/org/jitsi/impl/neomedia/stats/BasicReceiveStreamStatsImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        // Now check for lost packets.
+        int diff = RTPUtils.sequenceNumberDiff(seq, highestSeq);
+        if (diff <= 0)
+        {
+            // RFC3550 says that all packets should be counted as received.
+            // However, we want to *not* count retransmitted packets as received,
+            // otherwise the calculated loss rate will be close to zero as long
+            // as all missing packets are requested and retransmitted.
+            // Here we differentiate between packets received out of order and
+            // those that were retransmitted.
```
gpolitis: I think it is worth mentioning that the proper solution would be to enable RTX and detect losses before the de-RTX-fication.

gpolitis:
Hey @bgrozev, this is a significant amount of work and I'm trying to wrap my head around it. One thing that I find confusing is that many of the new classes have a corresponding removed class. For example:

AbstractMediaStreamSSRCStats-> AbstractBasicStreamStats
MediaStreamSSRCStats->BasicStreamStats
MediaStreamReceivedSSRCStats-> BasicReceiveStreamStatsImpl, BasicReceiveStreamStats
MediaStreamSentSSRCStats->BasicSendStreamStatsImpl, BasicSendStreamStats

Unless I’m missing something, is it impossible to keep the old class names?

bgrozev:
> Unless I’m missing something, is it impossible to keep the old class names?

Sure, but I prefer to keep the new names because:
1. The implementations are almost entirely new, so this wouldn't improve readability much
2. The API will still be incompatible with the old one
3. I think the new names (and package) are more appropriate

gpolitis:
1. The implementations are indeed almost entirely new, but the functionality/semantics isn't. And new implementation shouldn't mean we need to change the name.

2. Is it? https://github.com/jitsi/jitsi-videobridge/pull/284/files makes me think of the opposite. For example, the changes in CallStatsConferenceStatsHandler.java can be completely removed if we keep the existing naming.
 
3. The appropriateness of the names is debatable; I actually find the new names confusing. "stream" is already an overloaded term. I would rather see something that incorporates the term "track", like SendTrackStats or ReceiveTrackStats because that's what we really care about, i.e. we don't care about statistics on a per SSRC level (RTX or FEC ssrcs are irrelevant for instance). What we care about is statistics at the track level, that, by chance, it can have an associated RTX/FEC ssrc.

bgrozev:
> The implementations are indeed almost entirely new, but the functionality/semantics isn't. And new implementation shouldn't mean we need to change the name.

My point it that changing the names back won't make your job as a reviewer easier.

> Is it?

Yes, unless we also bring back the old inconsistent method names and move the accessor methods back to the MediaStreamStats interface (in which case we would have to move part of the implementation from MediaStreamStats2Impl to MediaStreamStatsImpl).


> The appropriateness of the names is debatable

Agreed.

> I would rather see something that incorporates the term "track", like SendTrackStats or ReceiveTrackStats because that's what we really care about

Sure, but this is not keeping the old class names. 

> we don't care about statistics on a per SSRC level

One more reason to not keep the old names :)

> What we care about is statistics at the track level, that, by chance, it can have an associated RTX/FEC ssrc.

We also use the same interfaces for the aggregated stats. But I like the "track" idea: TrackStats, SendTrackStats, ReceiveTrackStats -- gets rid of the word "basic" which is just superfluous. Do you want me to use these names instead?

gpolitis:
> Do you want me to use these names instead?

Yes, let's do that. I agree that makes more sense than both the old naming and the new naming.

bgrozev:
And what do you think about just Stats, ReceiveStats, SendStats? I think I prefer this to using "track" (which has other meanings, including one in FMJ)

gpolitis:
jenkins it's ok to test

PR #193, opened 2016-08-23
RTP uplifting frame reordering fixes

gpolitis:
I was analyzing a pcap file and I noticed re-order frames which the bridge mis-handles and causes it to break the RTP stream.

Say you send frames A B C and each is comprised of 3 RTP packets (3 is not important, it can be N). So the sender emits A1, A2, A3, B1, B2, B3, C1, C2, C3. If the bridge receives something like this A1, A2, A3, C1, B1, B2, B3, C2, C3 it will break the RTP stream because it will uplift the B frame to Timestamp_C+1.

tl;dr; the RTP uplifting logic can cope with some packet re-ordering, but it doesn’t expect whole frames to be re-ordered. I’ve confirmed that it can cause long freezes; I’m not sure about scrambling but I find it much less likely.


ibauersachs:
I might have missed some discussion, but the change int -> long in the SSRCs changes a lot of public API. I understand the motivation (and hate once again Java for not having uints), but is this really necessary? At some point much care was taken to NOT use longs and things refactored from long to int...

gpolitis:
@ibauersachs thanks for chiming in. I'm afraid of regressions so there's a good chance I'll take that back. I'm not aware of any decision to use ints instead of longs (fmj favors ints for sure) and it seems we're in favor of longs in LJ (at least in the past couple of years). I have reviewed a lot of PRs from @bgrozev and he's using them whenever possible. cc @lyubomir

gpolitis:
jenkins it's ok to test

src/org/jitsi/impl/neomedia/transform/rewriting/SsrcRewriter.java
```diff
@@ -106,6 +105,11 @@ protected boolean removeEldestEntry(Map.Entry eldest) {
         }
     };
 
+    /*
+     * 
+     */
+    private TimestampEntry maxSourceTsEntry;
```
damencho: Missing javadoc :)

src/org/jitsi/impl/neomedia/transform/rewriting/SsrcRewriter.java
```diff
@@ -287,7 +364,8 @@ void rewriteTimestamp(RawPacket p)
                     + " to " + newValue);
             }
 
-            tsHistory.put(oldValue, new TimestampEntry(now, newValue));
+            tsHistory.put(oldValue, new TimestampEntry(now, oldValue, newValue));
+            maxSourceTsEntry = new TimestampEntry(now, oldValue, newValue);
         }
```
damencho: Duplicate object creation.

src/org/jitsi/impl/neomedia/transform/DiscardTransformEngine.java
```diff
@@ -37,12 +39,33 @@
...
      * A map of source ssrc to {@link ResumableStreamRewriter}.
      */
     private final Map<Long, ResumableStreamRewriter> ssrcToRewriter
         = new HashMap<>();
 
     /**
+     *
+     */
+    private final MediaStream stream;
+
```
damencho: Missing javadoc.

src/org/jitsi/impl/neomedia/transform/DiscardTransformEngine.java
```diff
@@ -37,12 +39,33 @@
...
+    /**
+     *
+     * @param stream
+     */
+    public DiscardTransformEngine(MediaStream stream)
+    {
```
damencho: Missing javadoc.

src/org/jitsi/impl/neomedia/rtcp/RemoteClockEstimator.java
```diff
@@ -128,11 +140,15 @@ public void update(byte[] buf, int off, int len)
             }
         }
 
+
+        logger.debug("Updating the remote clock ssrc=" + ssrc
+            + ", systemTimeMs=" + systemTimeMs
+            + ", rtpTimestamp=" + rtptimestamp);
+
```
damencho: Maybe add if (logger.isDebugEnabled()) to avoid string creation.

src/org/jitsi/impl/neomedia/rtp/ResumableStreamRewriter.java
```diff
@@ -52,6 +52,8 @@
      */
     long highestTimestampSent = -1;
 
+    private int cycles = 0;
+
```
damencho: javadoc missing

PR #196, opened 2016-08-31
Sync fixes

bgrozev:


src/org/jitsi/impl/configuration/ConfigurationServiceImpl.java
```diff
@@ -102,7 +103,8 @@
      * and attempts to do so will simply be ignored.
      * @see #defaultProperties
      */
-    private Map<String, String> immutableDefaultProperties = new HashMap<>();
+    private Map<String, String> immutableDefaultProperties
+        = new ConcurrentHashMap<>();
```
lyubomir: That seems unnecessary because immutableDefaultProperties is modified only during the construction of ConfigurationServiceImpl.

src/org/jitsi/impl/configuration/ConfigurationServiceImpl.java
```diff
@@ -111,7 +113,8 @@
      * <tt>setProperty()</tt> methods. Still, re-setting one of these properties
      * to <tt>null</tt> would cause for its initial value to be restored.
      */
-    private Map<String, String> defaultProperties = new HashMap<>();
+    private Map<String, String> defaultProperties
+        = new ConcurrentHashMap<>();
```
lyubomir: Like immutableDefaultProperties, defaultProperties appears to be modified only during the construction of ConfiguratinoServiceImpl and, consequently, it seems unnecessary to use a ConcurrentHashMap.

src/org/jitsi/impl/neomedia/MediaUtils.java
```diff
@@ -53,7 +54,7 @@
      * defined in RFC 3551.
      */
     private static final Map<String, String> jmfEncodingToEncodings
-        = new HashMap<String, String>();
+        = new ConcurrentHashMap<>();
```
lyubomir: That seems unnecessary because jmfEncodingToEncodings is modified only during the static constructor of MediaUtils.

src/org/jitsi/impl/neomedia/MediaUtils.java
```diff
@@ -86,7 +87,7 @@
      */
     private static final Map<String, MediaFormat[]>
         rtpPayloadTypeStrToMediaFormats
-            = new HashMap<String, MediaFormat[]>();
+            = new ConcurrentHashMap<>();
```
lyubomir: Like jmfEncodingToEncodings, rtpPayloadTypeStrToMediaFormats appears to be modified only during the static constructor of MediaUtils and, consequently, it seems unnecessary to use a ConcurrentHashMap.

src/org/jitsi/impl/neomedia/MediaUtils.java
```diff
@@ -366,12 +367,20 @@
...
                 if (MediaType.AUDIO.equals(mediaFormat.getMediaType()))
+                {
                     audioMediaFormats.add(mediaFormat);
+                }
+            }
+        }
         for (MediaFormat mediaFormat : rtpPayloadTypelessMediaFormats)
+        {
             if (MediaType.AUDIO.equals(mediaFormat.getMediaType()))
                 audioMediaFormats.add(mediaFormat);
```
lyubomir: It looks inconsistent to me to add all {} in the previous loop and not add all in the next loop.

src/org/jitsi/impl/neomedia/format/MediaFormatImpl.java
```diff
@@ -255,11 +256,11 @@ protected MediaFormatImpl(
         this.formatParameters
             = ((formatParameters == null) || formatParameters.isEmpty())
                 ? EMPTY_FORMAT_PARAMETERS
-                : new HashMap<String, String>(formatParameters);
+                : new ConcurrentHashMap<>(formatParameters);
```
lyubomir: That seems unnecessary because the value of the fieldParameters is never modified in and exposed out of MediaFormatImpl.

src/org/jitsi/impl/neomedia/format/MediaFormatImpl.java
```diff
@@ -255,11 +256,11 @@ protected MediaFormatImpl(
...
         this.advancedAttributes
             = ((advancedAttributes == null) || advancedAttributes.isEmpty())
                 ? EMPTY_FORMAT_PARAMETERS
-                : new HashMap<String, String>(advancedAttributes);
+                : new ConcurrentHashMap<>(advancedAttributes);
```
lyubomir: The comment on formatParameters applies to advancedAttributes as well.

src/org/jitsi/impl/neomedia/format/ParameterizedVideoFormat.java
```diff
@@ -64,7 +65,7 @@ public ParameterizedVideoFormat(
         this.fmtps
             = ((fmtps == null) || fmtps.isEmpty())
                 ? MediaFormatImpl.EMPTY_FORMAT_PARAMETERS
-                : new HashMap<String, String>(fmtps);
+                : new ConcurrentHashMap<>(fmtps);
```
lyubomir: The Map instance stored in the field fmtps is never modified in and exposed out of ParameterizedVideoFormat.

src/org/jitsi/impl/neomedia/format/ParameterizedVideoFormat.java
```diff
@@ -145,7 +146,7 @@ protected void copy(Format f)
             fmtps
                 = ((pvfFmtps == null) || pvfFmtps.isEmpty())
                     ? MediaFormatImpl.EMPTY_FORMAT_PARAMETERS
-                    : new HashMap<String, String>(pvfFmtps);
+                    : new ConcurrentHashMap<>(pvfFmtps);
```
lyubomir: The same as above. The use of ConcurrentHashMap is unnecessary.

src/org/jitsi/impl/neomedia/recording/SynchronizerImpl.java
```diff
@@ -457,18 +449,15 @@ private int getPacketType(RawPacket pkt)
      */
     void removeMapping(long ssrc)
     {
-        if (ssrcs.containsKey(ssrc))
+        synchronized (ssrcs)
         {
-            synchronized (ssrcs)
+            SSRCDesc ssrcDesc = ssrcs.get(ssrc);
```
lyubomir: I'm afraid I don't get it. In removeMapping you protected the read with a synchronized so I guess you want reads to be protected. But then you left the read in getLocalTime unprotected so I'm confused now about what you want.

src/org/jitsi/impl/neomedia/recording/SynchronizerImpl.java
```diff
@@ -324,21 +321,16 @@ private SSRCDesc getSSRCDesc(long ssrc)
      */
     private Endpoint getEndpoint(String endpointId)
     {
-        Endpoint endpoint = endpoints.get(endpointId);
-        if (endpoint == null)
+        synchronized (endpoints)
         {
-            synchronized (endpoints)
+            Endpoint endpoint = endpoints.get(endpointId);
```
lyubomir: Unlike ssrcs, there's no protected block that does only reading so I don't think that you want the reads protected. Anyway, I'd like to mention getLocalTime again because there's an unprotected read from endpoints. Some analysis tools would flag such a synchronization as inconsistent.

src/org/jitsi/impl/configuration/ConfigurationServiceImpl.java
bgrozev: Thanks, Lyubo! I was trying to stay on the safe side and did not always follow all the usages. I'll remove the commits that change ConfigurationServiceImpl, MediaServiceImpl, MediaFormatImpl and ParameterizedVideoFormat.

src/org/jitsi/impl/neomedia/recording/SynchronizerImpl.java
bgrozev: You are right, the read doesn't need to be protected here. I simply refactored the existing code. Will remove it.

src/org/jitsi/impl/neomedia/rtp/StreamRTPManager.java
```diff
@@ -168,14 +171,17 @@ public void putResumableStreamRewriter(
     public ResumableStreamRewriter getResumableStreamRewriter(Long ssrc,
                                                               boolean create)
     {
-        ResumableStreamRewriter rewriter = ssrcToRewriter.get(ssrc);
-        if (rewriter == null && create)
+        synchronized (ssrcToRewriter)
```
gpolitis: I'm pretty sure there's only a single thread accessing the map. The documentation is not clear about, sorry about that. Would you like me to add a note explaining this and why it is the case?

src/org/jitsi/impl/neomedia/recording/SynchronizerImpl.java
bgrozev: Yeah, I'm fine with those unprotected reads. Protecting the single read of "ssrcs" was a mistake.

src/org/jitsi/impl/neomedia/transform/rewriting/SsrcGroupRewriter.java
```diff
@@ -66,7 +68,8 @@
      * A map of SSRCs to <tt>SsrcRewriter</tt>. Each SSRC that we rewrite in
      * this group rewriter has its own rewriter.
      */
-    private final Map<Integer, SsrcRewriter> rewriters = new HashMap<>();
+    private final Map<Integer, SsrcRewriter> rewriters
+        = new ConcurrentHashMap<>();
```
gpolitis: Same thing here, the `rewriters` object is accessed by a single thread at the time. We should be clear about that tho, to avoid confusion in the future.

src/org/jitsi/impl/neomedia/transform/DiscardTransformEngine.java
```diff
@@ -63,11 +63,16 @@ public RawPacket reverseTransform(RawPacket pkt)
                 = (pkt.getFlags() & Buffer.FLAG_DISCARD) == Buffer.FLAG_DISCARD;
 
             long ssrc = pkt.getSSRCAsLong();
-            ResumableStreamRewriter rewriter = ssrcToRewriter.get(ssrc);
-            if (rewriter == null)
+
+            ResumableStreamRewriter rewriter;
+            synchronized(ssrcToRewriter)
```
gpolitis: Same thing here, the `ssrcToRewriter` object is accessed by a single thread at the time. We should be clear about that tho, to avoid confusion in the future.

src/org/jitsi/impl/neomedia/rtp/StreamRTPManager.java
bgrozev: I would rather synchronize the block than depend on the fact that currently it is only used by one thread (the method is public and used in jitsi-videobridge).

src/org/jitsi/impl/neomedia/transform/rewriting/SsrcGroupRewriter.java
bgrozev: OK, will remove the commit.

src/org/jitsi/impl/neomedia/transform/DiscardTransformEngine.java
bgrozev: Are you sure about this? It is accessed by both the RTP and RTCP transformers, which IIRC are used by different threads.

src/org/jitsi/impl/neomedia/rtp/StreamRTPManager.java
gpolitis: That sounds sensible, but the method is used for every incoming RTP/RTCP packet, so I'd rather avoid introducing synchronization for obvious reasons. The JVB accesses this method sequentially, before the OutputDataStreamImpl.

bgrozev: Personally, I am more worried about someone wasting person time debugging a race condition months from now when the code has changed, than about the unknown gain in performance. But I'll remove the commit.

src/org/jitsi/impl/neomedia/transform/DiscardTransformEngine.java
gpolitis: (facepalm) You're absolutely right, we need that there.

src/org/jitsi/impl/neomedia/rtp/StreamRTPManager.java
bgrozev: Since this commit last week the map is accessed from the RTPConnectorOutputStream thread:
https://github.com/jitsi/libjitsi/commit/87f71421e59531e8a2382fcb398d3688227b0f65

bgrozev:
Moved to #197

PR #198, opened 2016-09-16
refactor(mediastream): Moves the isKeyFrame method to VideoMediaStream.

gpolitis:
Do _not_ merge yet. OTOH I don't like the idea of having to cast so frequently, but OTOH the `isKeyFrame` belongs to the `VideoMediaStream` interface.


gpolitis:
Closing as obsolete.

PR #208, opened 2016-10-28
Makes VP8 default video codec.

damencho:


brianh5:
looks like this rips out all the h264 support...what's the plan there? or maybe i'm misinterpreting

damencho:
oops sorry. will fix and resubmit

Issue #212, opened 2016-11-15
Incorrect assignment in InterArrival.java hinders bwe


brianh5:
Been tracking down some bad bwe in poor uplink scenarios (particularly with long delay links) and was finding that jvb wasn't reacting as well to delay as p2p in the same conditions.  Was finally able to track it down to a c++->java porting error in InterArrival::computeDeltas.  In that function the values for the current and previous timestamp groups are tracked, but when we start a new group and want to update the previous, it does an assignment...which is a copy in the c++ code but ends up being a reference assignment in java, so the 2 variables end up pointing to the same reference, so the timestamps always match.  After fixing that I got an accurate bwe for the scenario I was testing.  Will get a patch up.

gpolitis:
Thank you @brianh5, nice catch.

PR #213, opened 2016-11-15
Fix a reference assignment of prevTimestampGroup to correctly be a copy

brianh5:
https://github.com/jitsi/libjitsi/issues/212


bgrozev:
jenkins add to whitelist

PR #231, opened 2017-01-03
Fix the jitsi timestamp diff calculation error

EthanLin2015:
The original timestamp diff calculation by webrtc is using uint32, while in libjitsi they are using long instead, which allows negative value and will return incorrect result when calculating packetOrder.


bbaldino:
@gpolitis @bgrozev looks fine to me, but maybe you want to comment.  also i don't remember what we did on the CLA for highfive...if it was still individuals or if i got around to doing the corporate one.

ibauersachs:
Please watch the code style (braces in the first method). And I don't really like making methods more visible just to test them... I don't know that area of the code, but isn't there an already public method that can be called with varying parameters instead?

bbaldino:
agree with the visibility comment but it's a shame to not be able to test any methods that should be private, and testing a lower method from a higher level takes away a lot of the benefits of unit testing, in my opinion.

these tests in particular though seem liked they were more important to use during development as opposed to keeping them around to make sure the method doesn't get broken (as it's unlikely to see many changes i'd think/hope), so ethan how about changing the method back to private and removing the test?

ibauersachs:
Depends on how you define unit testing. Most often it's said that private methods shouldn't be tested as they are, well, private. In this particular case it shouldn't be too difficult to test with `computeDeltas`. And if it isn't possible to test that, then the change would be irrelevant.

But: the unit tests for libjitsi are disabled on Jenkins anyway... :(

EthanLin2015:
Thanks for the comment. I've fixed the code style. About the unit test, my original thought was to avoid the relatively complex logic in the ```computeDeltas``` method and focusing on the lower method. I added a separate class which contains the two methods for unit testing/ calling by the original methods in the updated commit. Let me know if it is feasible.

bbaldino:
cool, seems like this makes much more sense.  seem ok to you @ibauersachs ?

ibauersachs:
Yes. It would be nice to have complete Javadoc on all of those methods though.

PR #236, opened 2017-01-07
Update ice4j to 2.0-20161221.230043-4

ibauersachs:


bgrozev:
This failed because we haven't yet updated videobridge (and tests always run with latest master for everything). Will re-run in a bit.

bgrozev:
Jenkins ok to test

PR #237, opened 2017-01-10
updating BasicRTCPTerminationStrategy.java 

FearTheBadger:
to include a maximum bitrate for an RTCPREMBPacket


bgrozev:
Did you mean to limit this to 1Mbps? I think that's way too low.

FearTheBadger:
Fair, I'm going to pull back this request and refactor a little to make it so the max bitrate can be customizable.
As a default what do you feel is an acceptable max bitrate?

bgrozev:
We never needed a max here, so I'd prefer to keep the default unlimited. What is your use-case for limiting this?

PR #240, opened 2017-01-17
[stats] Record rate control statistics.

imedwei:
Created a mechanism to allow the flexible creation of counters to record the behavior of the bitrate estimators.

Not sure if you want this, but having a way to quickly instrument the rate controller with counters made it much easier to track down the rate control problem outlined in: https://github.com/jitsi/libjitsi/pull/239

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -193,6 +193,11 @@ public class MediaStreamImpl
     private MediaStreamStats2Impl mediaStreamStatsImpl;
 
     /**
+     * Contains statistics about this <tt>MediaStreamImpl</tt> used for telemetry and debugging.
+     */
+    private final StatisticsTable statisticsTable;
+
+    /**
      * The indicator which determines whether this <tt>MediaStream</tt> is set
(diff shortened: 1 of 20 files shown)


bbaldino:
Hey Ed, errors are:

02:11:19 [ERROR] /home/jenkins/jicofo/src/test/java/mock/media/MockMediaStream.java:[37,8] mock.media.MockMediaStream is not abstract and does not override abstract method getStatisticsTable() in org.jitsi.service.neomedia.MediaStream
02:11:19 [ERROR] /home/jenkins/jicofo/src/test/java/mock/media/MockVideoMediaStream.java:[35,8] mock.media.MockVideoMediaStream is not abstract and does not override abstract method getStatisticsTable() in org.jitsi.service.neomedia.MediaStream

imedwei:
Oh I see, the interface change affected a component we don't use.

How do I pull request both components atomically?
How would you recommend we build/test the jitsi components?

Or perhaps this may be all unnecessary if we find that the interface need not change?

imedwei:
@bbaldino Do you have a suggestion how I can simultaneously submit patches for both repositories?

bbaldino:
@imedwei pretty sure if you use the same branch name for both it'll pull from both for the checks

PR #241, opened 2017-01-24
[h264] Fix infinite loop in Single-Time Aggregation Packet type A (STAP-A) validation.

imedwei:
Due to sign extension, the 16 bit NAL unit length may be read as a negative value
    which may cause the loop to either iterate infinitely or falsely abort the loop and
    mark the STAP-A as valid.

    Test Plan: Unit tests that exercises the edge cases.


imedwei:
@bbaldino You might be interested in this one.

bbaldino:
this is the infinite loop one from a while back, right?

imedwei:
Yes, that's right.

bbaldino:
@gpolitis @bgrozev this was an issue with the h264 code we submitted a while back that was found a couple months ago.  it's h264-specific, so nothing jitsi would be seeing in any prod environment currently, but it's been running in highfive prod for a while now.  any concerns?

bgrozev:
Is taking into account "off" the only fix here, and the rest is just refactoring? And this shouldn't affect the Jitsi client, I suppose?
LGTM

PR #242, opened 2017-01-24
Add org.jitsi.impl.neomedia.RawPacket.DISCARD_CONTRIBUTING_SOURCES config property

debugmaster:
This property is useful when a SIP Provider does not support the list of contributing sources in its SIP infrastructure. If one is having annoying noise issues, this property may solve it when it is set to true.

### DIFF
--- a/src/org/jitsi/impl/neomedia/RawPacket.java
+++ b/src/org/jitsi/impl/neomedia/RawPacket.java
@@ -15,6 +15,10 @@
  */
 package org.jitsi.impl.neomedia;
 
+import org.jitsi.service.configuration.*;
+import org.jitsi.service.libjitsi.*;
+import org.jitsi.service.neomedia.*;
+import org.jitsi.util.*;
 
 import org.jitsi.service.neomedia.*;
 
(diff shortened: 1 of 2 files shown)


debugmaster:
I've signed the CLA

bgrozev:
Thank you for the contribution. However, I don't think the basic RawPacket class is the right place for such logic. It should go to MediaStreamImpl or CsrcTransformEngine, or whatever code you are using which calls setCsrcList()

debugmaster:
I reverted all changes from RawPakcet and moved to CsrcTranformEngine, which is the only class that calls setCsrcList().

I added the property verification to [this line](https://github.com/jitsi/libjitsi/pull/242/files/f8ea91cf1c5b8d200d9c4ccf053364e7d3e5a850..66bd7c3cc46e55de03e7f1d4355d057998f11183#diff-f1d6d3bb587859167d31a6f97ca08d82R313). Tell me if I should place it somewhere else.

bgrozev:
Jenkins, test this please

bgrozev:
@debugmaster LGTM

PR #244, opened 2017-01-30
Remove fir pli parsing

gpolitis:


src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -1067,6 +1061,12 @@ private TransformEngineChain createTransformEngineChain()
             engineChain.add(mediaStreamTrackReceiver);
         }
 
+        // Debug
+        debugTransformEngine
```
bbaldino: is this on the ingress or egress side?

gpolitis: It's in both the ingress and the egress.

bbaldino: so this mean the debug engine runs (close to) first on the ingress side, and (close to) last on the egress, right?

PR #248, opened 2017-01-31
feat: Disables codecs which depend on ffmpeg by default.

bgrozev:


ibauersachs:
Wouldn't it be better to have a mechanism to lookup the available codecs, e.g. via JNDI or a similar OSGi mechanism?

bgrozev:
I don't know if that would work with codecs which require a native library.

ibauersachs:
Why not? I'm just referring to something that reads the list of these currently hard-coded class names from e.g. a manifest and then loads them in exactly the same way as now.

bgrozev:
> Why not? I'm just referring to something that reads the list of these currently hard-coded class names from e.g. a manifest and then loads them in exactly the same way as now.

Sorry, I missed this comment. Sure, I don't see why that wouldn't work. But I don't see a significant benefit either. Are you against merging this?

bgrozev:
Jenkins test this please

ibauersachs:
I would just prefer to make things dynamic once, not adding specifics for one library.
If that's not possible due to time constraints, go ahead with this PR.

PR #249, opened 2017-01-31
Supports the framemarkings extension

bgrozev:
for detecting keyframes.


src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * Initializes a new {@link ByteArrayBufferImpl} instance.
+     * @param buffer
+     * @param offset
+     * @param length
+     */
+    public ByteArrayBufferImpl(byte[] buffer, int offset, int length)
+    {
+        this.buffer = Objects.requireNonNull(buffer, "buffer");
+        if (offset + length > buffer.length)
```
gpolitis: do we care to check for the length to be positive here?

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
```diff
@@ -0,0 +1,145 @@
...
+        return length;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void setLength(int length)
+    {
+        if (offset + length > buffer.length)
```
gpolitis: do we care to check the length to be positive here?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -3678,10 +3688,47 @@ public boolean isEndOfFrame(byte[] buf, int off, int len)
...
-    * {@inheritDoc}
-    */
+     * {@inheritDoc}
+     * </p>
+     * This is absolutely terrible, but we need a RawPacket and the method is
+     * used from RTPTranslator, which doesn't work with RawPacket.
+     */
     public boolean isKeyFrame(byte[] buf, int off, int len)
     {
+        return isKeyFrame(new RawPacket(buf, off, len));
```
gpolitis: instead of creating a new object, can we make this the main method and pass pkt.getBuffer() etc from the other method?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -3678,10 +3688,47 @@ public boolean isEndOfFrame(byte[] buf, int off, int len)
...
+     */
+    public boolean isKeyFrame(RawPacket pkt)
+    {
+        byte[] buf = pkt.getBuffer();
+        int off = pkt.getOffset();
+        int len = pkt.getLength();
+
+        if (frameMarkingsExtensionId != -1)
+        {
+            RawPacket.HeaderExtensions hes = pkt.getHeaderExtensions();
```
gpolitis: in order for the above recommendation to work, the getHeaderExtensions() can be made static.

src/org/jitsi/service/neomedia/ByteArrayBuffer.java
```diff
@@ -59,6 +59,13 @@
     void setOffset(int off);
 
     /**
+     * Sets the offset and length fields of this {@link ByteArrayBuffer}.
+     * @param off the offset to set.
+     * @param len the length to set.
+     */
+    void setOffsetLength(int off, int len);
```
gpolitis: Instead of forcing all inheritors to implement this method, can we make it an implementation method and use it from the implementation instance?

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
bgrozev: Well, the way I use it in RawPacket is I initialize it with length 0 and then update it.

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
bgrozev: I would rather we accept length of 0, but I guess I can make changes to avoid it. Let me know which one you prefer.

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
bgrozev: getHeaderExtensions uses an instance field, and we would have to create a new instance there if it were to be made static.
What is the motivation for making the isKeyFrame(byte[], int, int) the main method?

src/org/jitsi/service/neomedia/ByteArrayBuffer.java
bgrozev: Yep, will do.

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
gpolitis: You're right, length of 0 should be valid.

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
gpolitis: > getHeaderExtensions uses an instance field, and we would have to create a new instance there if it were to be made static.
oh, right, I see. Then it's fine the way it is.

> What is the motivation for making the isKeyFrame(byte[], int, int) the main method?
To not have to create a new `RawPacket`, if you have a (buf, off, len) triplet, but this is just a negligible detail.

gpolitis:
Could you please add the new `isKeyFrame` method to `AbstractMediaStream` so that the smoke tests can run?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -3678,10 +3688,47 @@ public boolean isEndOfFrame(byte[] buf, int off, int len)
...
     public boolean isKeyFrame(byte[] buf, int off, int len)
     {
+        return isKeyFrame(new RawPacket(buf, off, len));
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public boolean isKeyFrame(RawPacket pkt)
+    {
```
gpolitis: Are we sure a `null` never reaches this point?

src/org/jitsi/impl/neomedia/rtp/FrameMarkingHeaderExtension.java
```diff
@@ -0,0 +1,65 @@
...
+    private static byte KF_MASK = (byte) (S_BIT | I_BIT);
+
+    /**
+     * @return true if the extension contained in the given buffer indicates
+     * that the corresponding RTP packet is the first packet of a keyframe (i.e.
+     * the S and I bits are set).
+     */
+    public static boolean isKeyframe(ByteArrayBuffer baf)
+    {
+        if (baf.getLength() < 2)
```
gpolitis: Are we sure a `null` never reaches this point?

bgrozev:
I added a PR in jicofo (adding to MockMediaStream instead, because RawPacket is in impl, and AbstractMediaStream is in service).

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
bgrozev: Added a check.

src/org/jitsi/impl/neomedia/rtp/FrameMarkingHeaderExtension.java
bgrozev: Added a check.

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
gpolitis: Could you please add that `length > -1` check here? I expect this base class to be used in various places so we need to add proper argument checks.

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
gpolitis: I expect this to be a commonly used base class for specialized `ByteArrayBuffer`s. Could you please add the positive or 0 check that we discussed bellow?

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void setOffset(int offset)
+    {
+        if (offset + length > buffer.length)
+        {
+            throw new IllegalArgumentException("length");
```
gpolitis: typo

gpolitis: And maybe a check for negative offsets?

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
```diff
@@ -0,0 +1,149 @@
...
+        this.length = length;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isInvalid()
+    {
+        return false;
```
gpolitis: To complete my review of negligible comments, do you think we could change this to `buf != null && buf.length >= off + len && len > -1`

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
bgrozev: Yes, adding.

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
bgrozev: Added.

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
bgrozev: Fixed

src/org/jitsi/impl/neomedia/ByteArrayBufferImpl.java
bgrozev: These conditions are enforced by the public API, so it would be equivalent to "return false", unless there's a bug somewhere. Will add it if you still think it should be there.

gpolitis: You're right, thanks :+1:

src/org/jitsi/impl/neomedia/RawPacket.java
```diff
@@ -678,8 +682,7 @@ public int getHeaderExtensionType()
             return 0;
 
         return
-            readUnsignedShortAsInt(
-                    offset + FIXED_HEADER_SIZE + getCsrcCount() * 4);
+            readUint16AsInt(offset + FIXED_HEADER_SIZE + getCsrcCount() * 4);
```
gpolitis: maybe we should rename this to readUshortAsInt?

PR #252, opened 2017-02-01
Add extra debug information for the sctp crash

bbaldino:


src/native/build.xml
```diff
@@ -1334,7 +1336,7 @@
             <linkerarg value="-lusrsctp" location="end" />
             <linkerarg value="-Wl,-Bdynamic" location="end" />
             <linkerarg value="-lpthread" location="end" />
-            <linkerarg value="-L${usrsctp}/usrsctplib/.libs" />
+            <linkerarg value="-L/home/brian/src/usrsctp/usrsctplib/.libs" />
```
bbaldino: oops, let me clean this up

bbaldino:
this is intended as a TEMPORARY commit, to be reverted once we get the crash sorted out

src/org/jitsi/sctp4j/Sctp.java
```diff
@@ -174,6 +174,7 @@ public static synchronized void init()
         //    return;
         if(!initialized)
         {
+            logger.error("Init'ing brian's patched usrsctp");
```
paweldomas: This came here by mistake right ;) ?

bbaldino: haha actually this is intentional :) i can remove the 'brian' part of it if we want, but i added some prints so we can know that the proper lib is loaded

damencho:
Jenkins, test this please.

PR #254, opened 2017-02-02
add extra checks for the memory addr whenever we assign the next ptr

bbaldino:


ibauersachs:
I realize this is for debugging, but please commit the source code changes along with the binary (or in the same pull request).

bbaldino:
@ibauersachs the usrsctp source isn't in this repo anywhere, though might not be a bad idea to do that one way or another

ibauersachs:
Hm, okay, I thought this is `src/native/sctp`. If you refer to a specific fork/commit of https://github.com/sctplab/usrsctp.git, please mention that somewhere in the commit (e.g. like this: sctplab/usrsctp@d5f916d2a42606a7660384e8cbc9a05a933815b4).

bbaldino:
it is `src/native/sctp` but we don't actually commit any source there.  i have been working off of a local fork of usrsctp and then pushing the lib changes.  i can push that...there's nothing pretty here though and i wouldn't recommend anyone for at this point--it's unfortunate that this has to go in at master at all but we're working around other limitation.

my repo is here with the changes form this update is here: https://github.com/bbaldino/usrsctp/commit/27124c4051ca731204b97600fec72a5e6822ccc1

PR #256, opened 2017-02-03
Adds support for OHB 

bgrozev:
(Original Header Block) RTP header extensions.

### DIFF
--- a/src/org/jitsi/impl/neomedia/transform/OriginalHeaderBlockTransformEngine.java
+++ b/src/org/jitsi/impl/neomedia/transform/OriginalHeaderBlockTransformEngine.java
@@ -54,7 +54,7 @@ public class OriginalHeaderBlockTransformEngine
     /**
      * The ID of the OHB RTP header extension, or -1 if it is not enabled.
      */
-    private byte extensionID = 8;
+    private byte extensionID = -1;
 
     /**
      * Initializes a new {@link OriginalHeaderBlockTransformEngine} instance.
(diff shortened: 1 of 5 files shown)


src/org/jitsi/impl/neomedia/transform/OriginalHeaderBlockTransformEngine.java
```diff
@@ -0,0 +1,232 @@
...
+        else if (pt && !seq && ts && ssrc)
+            return 9;
+        else if (!pt && seq && ts && ssrc)
+            return 10;
+        else if (pt && seq && ts && ssrc)
+            return 11;
+
+        // The above is exhaustive, but the compiler (and IDE) can't figure it
+        // out.
+        return -1;
```
bbaldino: is it better to throw here?

bgrozev: Yep, made the change.

gpolitis: If this is exhaustive, can't we change the last `else if` with an `else`?

src/org/jitsi/impl/neomedia/transform/OriginalHeaderBlockTransformEngine.java
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * Implements {@link SinglePacketTransformer#reverseTransform(RawPacket)}.
+     */
+    @Override
+    public RawPacket reverseTransform(RawPacket pkt)
+    {
+        if (extensionID != -1)
+        {
+            // TODO: check if an OHB ext already exists.
```
gpolitis: this should be an easy addition, but then again the only benefit here would be the removal of the TODO text..

bgrozev: Actually it wouldn't be a simple addition, because the semantics of an existing OHB extension are somewhat complicated (if one exists then the bridge is allowed to modify extensions which come after it, and not extensions which come before it).
https://tools.ietf.org/html/draft-ietf-perc-double-02#section-5.2

PR #259, opened 2017-02-06
Fix short to uint conversion

gpolitis:


src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -138,114 +138,106 @@ public MediaStreamTrackReceiver getMediaStreamTrackReceiver()
...
-
-        for (int i = encoding.getIndex() + 1; i < rtpEncodings.length; i++)
+        if (!frameDesc.isIndependent())
         {
-            RTPEncodingDesc enc = rtpEncodings[i];
-            FrameDesc lastReceivedFrame = enc.getLastReceivedFrame();
+            RTPEncodingDesc encoding = frameDesc.getRTPEncoding();
+
+            // When we suspect that a stream is suspended, we send a FIR to
+            // downscale receivers that are receiving the suspended stream.
```
bbaldino: i'm confused by this comment.  we're sending FIR to the stream "below" the one that just got suspended so receivers can switch to it, right?  maybe something like:

`When we suspect that a stream is suspended, we send an FIR to the sender so we can send a different stream to its receivers` ?

bbaldino:
might be good to rename the PR, the short -> uint conversion seems to be a minor aspect of this change :)

gpolitis:
I pushed a bit more than what I wanted. The changes in the stream suspension detection should be here:  https://github.com/jitsi/libjitsi/pull/258

Please do not merge this like this.

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
gpolitis: thanks Brian, I'll change this in #258.

bgrozev:
lgtm

bbaldino:
i actually like the other naming better (where we explicitly state the amount of bits we were reading) instead of this...why the more vague naming?

bgrozev:
> i actually like the other naming better (where we explicitly state the amount of bits we were reading) instead of this...why the more vague naming?

I agree, but I'm also fine with the new names.

gpolitis:
I'll delete the styling commits.

bgrozev:
Tests passed here: http://ci.jitsi.org/view/pr-testing/job/libjitsi-PR-test/356/

PR #260, opened 2017-02-07
Move rawpacket to service

damencho:


src/org/jitsi/service/neomedia/RawPacket.java
```diff
@@ -13,10 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jitsi.impl.neomedia;
+package org.jitsi.service.neomedia;
 
 
-import org.jitsi.service.neomedia.*;
+import org.jitsi.impl.neomedia.*;
```
ibauersachs: Stuff from .service. shouldn't reference anything inside .impl. to truly follow the service vs. impl packaging

damencho: Ups. Checking it now.

gpolitis: @damencho would it make sense to use `org.jitsi.service...ByteArrayBuffer`? or does that break stuff?

bgrozev:
LGTM

PR #268, opened 2017-02-21
remove chatty log for good pointer assignment

bbaldino:


paweldomas:
Is there a reference to the actual change in the source on Github ? Can we at least include a link here ?

bbaldino:
yup, good idea.  the code behind this binary is here: https://github.com/bbaldino/usrsctp/tree/fde2f44a0e428e7fdf5e361b108dd697a4f64648

it would be nice to put usrsctp as a submodule or something so we can keep this in sync in the future.

PR #273, opened 2017-02-28
Rtp hdrext padding

bgrozev:


bgrozev:
Please don't merge yet

bgrozev:
Jenkins ok to test

src/org/jitsi/service/neomedia/RawPacket.java
```diff
@@ -203,107 +203,202 @@ public static long getRTCPSSRC(byte[] buf, int off, int len)
...
+        // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+        // | A: unused | B: hdr + ext | C: payload | D: unused |
+        // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+        // And the regions have the following sizes:
+        // A: this.offset
+        // B: this.getHeaderLength()
+        // C: this.getPayloadLength()
+        // D: this.buffer.length - this.length - this.offset
+        // We will try to extend the packet so that it uses A and/or D if
+        // possible, in order to avoid allocating new memory.
```
gpolitis: this is cool!

bbaldino:
i don't really get overloading RawPacket with RTP-aware data like this, it seems to me like RawPacket should just worry about modeling a raw packet, and anything rtp-related should be in a separate class (which may leverage a RawPacket for the underlying buffer, for example).  i know RawPacket was already doing more before this diff, just seems like this is taking us further down that road.

bgrozev:
@bbaldino this PR merely re-implements an existing method. I agree about not mixing RTP, RTCP and pure buffer handling code, but I think at this point the class called "RawPacket" mostly handles RTP anyway. I think a way out of this would be to make RawPacket extend ByteArrayBuffer and start moving some generic buffer handling code to ByteArrayBuffer.

bbaldino:
thanks @bgrozev, didn't realize this wasn't adding anything new.  sounds like that might be a good plan for the future

src/org/jitsi/service/neomedia/RawPacket.java
```diff
@@ -203,107 +203,202 @@ public static long getRTCPSSRC(byte[] buf, int off, int len)
...
+
+        // This is an upper bound on the required length for the packet after
+        // the addition of the new extension. It is easier to calculate than
+        // the exact number, and is relatively close (it may be off by a few
+        // bytes due to padding)
+        int maxRequiredLength
+            = getLength()
+            + (extensionBit ? 0 : EXT_HEADER_SIZE)
+            + len + 1 /* header extension header */
+            + 3 /* padding */;
```
gpolitis: Do you think this still applies? (taken from CsrcTransformEngine.java)

    // calculate extension padding		
    int padLen = 4 - (len + 1) % 4;
    if (padLen == 4) padLen = 0;

src/org/jitsi/service/neomedia/RawPacket.java
```diff
@@ -203,107 +203,202 @@ public static long getRTCPSSRC(byte[] buf, int off, int len)
...
+
+
+        // This is an upper bound on the required length for the packet after
+        // the addition of the new extension. It is easier to calculate than
+        // the exact number, and is relatively close (it may be off by a few
+        // bytes due to padding)
+        int maxRequiredLength
+            = getLength()
+            + (extensionBit ? 0 : EXT_HEADER_SIZE)
+            + len + 1 /* header extension header */
```
gpolitis: header extension length? (typo)

src/org/jitsi/service/neomedia/RawPacket.java
bgrozev: Well, the thing is, how much padding we will add depends on the length of the existing extension and it's padding. And to find the padding of the existing extension we have to iterate over it. So instead of doing this, I use an upper bound of the padding which we are going to add, which is just 3.
The actual padding is handled below, in a way similar to the code that you cite (and it is RawPacket.addExtension which handles the padding, and not the classes which use it).

src/org/jitsi/service/neomedia/RawPacket.java
bgrozev: Will clarify

src/org/jitsi/service/neomedia/RawPacket.java
gpolitis: got it, thanks @bgrozev.

gpolitis:
LGTM

PR #275, opened 2017-03-01
Revert "fix: Don't count packets which failed to retransmit."

bgrozev:
This reverts commit 355069199c665eecbdba24b91d66d3b6659ba84a.


saghul:
For posterity (and for more pleasant git archeology expeditions) can we state why something was reverted in the commit log? 🙏

PR #276, opened 2017-03-02
Implements support for transport-cc

bgrozev:
Note that right now this implements little else than handling the configuration of transport-cc in a MediaStream. I'm just opening the PR now in order to have the tests run.


src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -0,0 +1,152 @@
...
+            if (extensionId != -1)
+            {
+                RawPacket.HeaderExtension ext
+                    = pkt.getHeaderExtension(extensionId);
+                if (ext == null)
+                {
+                    ext = pkt.addExtension(extensionId, 2);
+                }
+
+                int seq = outgoingSeq.getAndIncrement() & 0xffff;
```
bbaldino: i'm sure this is existing code, but it'd be nice to add a helper to do this masking for us (`...AsInt()` or something).  and is there no concern about a loss of data when masking off like this?

bbaldino:
i'll leave the full review to others, just a drive-by

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bgrozev: It's not existing code, but I don't understand what you mean.

bbaldino: why do we need the mask?

bgrozev: Because the on-the-wire format uses 16 bits, but the running counter is an AtomicInteger (that was easier to implement than keeping a counter which resets, because access to it needs to be synchronized).

bbaldino: ah, i understand now.  thanks.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,325 @@
...
+     *
+     * Note that this implementation is not optimized and might not always use
+     * the minimal possible number of bytes to describe a given set of packets.
+     * Specifically, it does take into account that sequence numbers wrap
+     * at 2^16 and fails to pack numbers close to 2^16 with those close to 0.
+     */
+    public RTCPTCCPacket(long senderSSRC, long sourceSSRC,
+                         TreeMap<Integer, Long> packets,
+                         byte fbPacketCount)
+    {
```
bbaldino: lots of nice comments but still feels like a logic of magic numbers/formulae in here.  i know this is kinda the core method though, so maybe this is as good as it can be.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,325 @@
...
+        this.packets = packets;
+    }
+
+    /**
+     *
+     * @param next
+     * @return
+     */
+    /*
+    public static Map<Integer, Long> getPackets(ByteArrayBuffer next)
```
bbaldino: not needed?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,325 @@
...
+    /**
+     * @return
+     */
+    synchronized public Map<Integer, Long> getPackets()
+    {
+        if (packets == null)
+        {
+            // parse this.fci as containing NACK entries and initialize
+            // this.packets
+            //packets = getPackets(new RawPacket(fci, 0, fci.length));
```
bbaldino: should this case be handled?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -50,12 +65,156 @@
...
+     * @param marked whether the RTP packet had the "marked" bit set.
+     */
+    private void packetReceived(int seq, boolean marked)
+    {
+        long now = System.currentTimeMillis();
+        synchronized (incomingSeqsSyncRoot)
+        {
+            if (incomingSeqs == null)
+            {
+                incomingSeqs = new TreeMap<>(RTPUtils.sequenceNumberComparator);
```
bbaldino: any reason not to do this at construction time?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -50,12 +65,156 @@
...
+    /**
+     * Notifies this instance that a data packet with a specific transport-wide
+     * sequence number was received on this transport channel.
+     * @param seq the transport-wide sequence number of the packet.
+     * @param marked whether the RTP packet had the "marked" bit set.
+     */
+    private void packetReceived(int seq, boolean marked)
+    {
+        long now = System.currentTimeMillis();
+        synchronized (incomingSeqsSyncRoot)
```
bbaldino: since this method just writes to `incomingSeqs` and the other method (`maybeSendRtcp`) just reads, why not use one of the thread-safe producer/consumer data structures instead of explicitly synchronizing?  i know we're doing a drop here if it's full, but i think there are structures which provide that behavior?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -149,4 +320,46 @@ private RTCPTransformer()
...
+     * use this {@link TransportCCEngine}.
+     * @param mediaStream the stream to remove.
+     */
+    public void removeMediaStream(MediaStream mediaStream)
+    {
+        synchronized (mediaStreams)
+        {
+            while(mediaStreams.remove(mediaStream))
+            {
+                // remove all
```
bbaldino: still TODO?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -149,4 +320,46 @@ private RTCPTransformer()
...
+            }
+        }
+    }
+
+    /**
+     * @return one of the {@link MediaStream} instances which use this
+     * {@link TransportCCEngine}, or null.
+     */
+    private MediaStream getMediaStream()
```
bbaldino: i'm curious...what does the use case end up being where jut getting the first one is sufficient?

src/org/jitsi/util/RTPUtils.java
```diff
@@ -69,6 +71,27 @@ public static int writeInt(byte[] buf, int off, int data)
...
 
     /**
+     * Writes the least significant 24 bits from the given integer into the
+     * given byte array at the given offset.
+     * @param buf the buffer into which to write.
+     * @param off the offset at which to write.
+     * @param data the integer to write.
+     * @return 3
+     */
+    public static int writeUint24(byte[] buf, int off, int data)
```
bbaldino: feels like it belongs somewhere more generic than 'RTPUtils' but not a big deal for now i guess

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
```diff
@@ -47,13 +48,76 @@
...
+        byte[] fci = fciBuffer.getBuffer();
+        int off = fciBuffer.getOffset(), len = fciBuffer.getLength();
+
+        for (int i = 0; i < (len / 4); i++)
+        {
+            int pid = (0xFF & fci[off + i * 4 + 0]) << 8 | (0xFF & fci[off + i * 4 + 1]);
+            lostPackets.add(pid);
+
+            // First byte of the BLP
+            for (int j = 0; j < 8; j++)
```
bgrozev: Added braces throughout the file.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,325 @@
...
+     * not.
+     *
+     * @param baf the {@link ByteArrayBuffer}
+     * @return true if the byte array buffer holds a NACK packet, otherwise
+     * false.
+     */
+    public static boolean isTCCPacket(ByteArrayBuffer baf)
+    {
+        int rc = RTCPHeaderUtils.getReportCount(baf);
+        return isRTPFBPacket(baf) && rc == FMT;
```
bgrozev: Changed.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Do you think splitting TreeMap<Integer, Long> with the given Comparator into a new class will help?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: To be implemented.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Yeah, I'm sorry I should have mentioned -- that part will be implemented shortly (I hope).

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bgrozev: Yeah, this executes multiple times, because once we send feedback we set it back to null.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bgrozev: maybeSendRtcp also sets it to null. We could make it clear the map instead, but that would be writing. So, I don't see right away how to reuse an existing structure.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bgrozev: The intention was to explain the existence of the while loop, will clarify.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bgrozev: With bundle we have more than one MediaStream instances which share a transport (and therefore a TransportCCEngine). We only use the MediaStream for the injectPacket method, which effectively writes to the same socket regardless of the MediaStream.

src/org/jitsi/util/RTPUtils.java
bgrozev: Agreed. We'd have to also move the other write* and read* methods.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bbaldino: which comparator is that?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bbaldino: `ArrayBlockingQueue` has an atomic `clear`, would that work?  Seems like clearing it (rather than setting to null and re-allocating a new one) might be nice as well?

bgrozev: That would work, but inserts might be expensive (if we stick to keeping it ordered the whole time).

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: The one in TransportCCEngine (which compares sequence numbers that wrap at 2^16)

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
```diff
@@ -46,49 +46,11 @@
...
-        }
-    };
-
-    /**
      * Stores the RTP payloads (VP8 payload descriptor stripped) from RTP packets
      * belonging to a single VP8 compressed frame.
      */
-    private SortedMap<Long, Container> data
-            = new TreeMap<Long, Container>(seqNumComparator);
+    private SortedMap<Integer, Container> data
```
bbaldino: does this need to be `Integer` (as opposed to `int`)?  thinking about all those Integer allocations from int conversions...

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
```diff
@@ -46,49 +46,11 @@
...
-                    return -1;
-                else
-                    return 1;
-            }
-        }
-    };
-
-    /**
      * Stores the RTP payloads (VP8 payload descriptor stripped) from RTP packets
      * belonging to a single VP8 compressed frame.
```
bbaldino: could we add in the doc what this maps?  E.g. "maps rtp sequence number to something something"

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
```diff
@@ -47,13 +48,84 @@
...
+     * not.
+     *
+     * @param baf the {@link ByteArrayBuffer}
+     * @return true if the byte array buffer holds a NACK packet, otherwise
+     * false.
+     */
+    public static boolean isNACKPacket(ByteArrayBuffer baf)
+    {
+        int rc = RTCPHeaderUtils.getReportCount(baf);
+        return isRTPFBPacket(baf) && rc == FMT;
```
bbaldino: i think we had talked about swapping the order of these so we can short-circuit?

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
```diff
@@ -47,13 +48,84 @@
...
+    }
+
+    /**
+     * @return the set of sequence numbers reported lost in a NACK packet
+     * represented by a {@link ByteArrayBuffer}.
+     * @param baf the NACK packet.
+     */
+    public static Collection<Integer> getLostPackets(ByteArrayBuffer baf)
```
bbaldino: wondering again about the need for `Integer`

bbaldino: EDIT: i see now that was another place, but maybe it'd be good here too

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,322 @@
...
+     * @return true if the byte array buffer holds a NACK packet, otherwise
+     * false.
+     */
+    public static boolean isTCCPacket(ByteArrayBuffer baf)
+    {
+        int rc = RTCPHeaderUtils.getReportCount(baf);
+        return rc == FMT && isRTPFBPacket(baf);
+    }
+
+    public static TreeMap<Integer, Long> getPackets(ByteArrayBuffer baf)
```
bbaldino: docs?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,322 @@
...
+    {
+        return getPacketsFci(getFCI(baf));
+    }
+
+    /**
+     *
+     * @param next
+     * @return
+     */
+    public static TreeMap<Integer, Long> getPacketsFci(ByteArrayBuffer fciBuffer)
```
bbaldino: docs

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bbaldino: hard to tell for me.  i was more referring to a lot of the raw numbers used in various equations below...i'll try to comment on the individual lines

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,322 @@
...
+        TreeSet<Map.Entry<Integer, Long>> sequenceNumbers
+            = (TreeSet) packets.entrySet();
+
+        Map.Entry<Integer, Long> first = sequenceNumbers.first();
+        int firstSeq = first.getKey();
+        Map.Entry<Integer, Long> last = sequenceNumbers.last();
+        int packetCount
+            = 1 + RTPUtils.sequenceNumberDiff(last.getKey(), firstSeq);
+
+        byte[] buf = new byte[(packetCount / 7 + 1) * 2 + 8];
```
bbaldino: what is `(packetCount / 7 + 1) * 2 + 8` doing? i think i get 'one byte per packet', but why double it?  and then there's an extra byte needed for some overhead or something?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,322 @@
...
+        byte[] deltas = new byte[packetCount * 2];
```
bbaldino: why do we double packetCount here?  i'm sure these are transport-cc implementation details, but maybe there's some rfc text we can copy or something.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,322 @@
...
+                buf[off] = 0;
+            }
+
+            int symbol;
+
+            int seq = (firstSeq + seqDelta) & 0xffff;
+            Long ts = packets.get(seq);
+            if (ts == null)
+            {
+                symbol = 0; // packet not received
```
bbaldino: constants for these symbols (packet not received, packet received, small delta, etc.) would make this more self-documenting i think

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -0,0 +1,365 @@
...
+ *
+ * @author Boris Grozev
+ */
+public class TransportCCEngine
+    implements TransformEngine
+{
+    /**
+     * The maximum number of received packets and their timeastamps to save.
+     */
+    private static final int MAX_INCOMING_PACKETS_HISTORY = 200;
```
bbaldino: should this be public now?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bbaldino: my bad, totally missed this was a treemap and was more than just a simple queue

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bbaldino: what's the reason to set this to null and re-allocate it as opposed to just clearing it and checking for empty?

src/org/jitsi/impl/neomedia/transform/OriginalHeaderBlockTransformEngine.java
```diff
@@ -225,15 +225,15 @@ public PacketTransformer getRTCPTransformer()
...
 
-        RTPUtils.writeShort(extensionBytes, 1, (short) pkt.getSequenceNumber());
-        RTPUtils.writeInt(extensionBytes, 3, (int) pkt.getTimestamp());
-        RTPUtils.writeInt(extensionBytes, 7, pkt.getSSRC());
+        byte[] buf = he.getBuffer();
+        int off = he.getOffset();
 
-
-        pkt.addExtension(extensionID, extensionBytes);
+        buf[off + 1] = pkt.getPayloadType();
```
bbaldino: why do we skip a byte?

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
bgrozev: Yeah, we can't store primitives in collections :(
That's probably why we see so many allocations in the first place.

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
bgrozev: Will fix, I didn't realize there was more than one place that has it.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Seven packets per packet status chunk which is two bytes long plus an additional 8 bytes for the fixed fields, obviously. :)
I think there's a bug in here, will fix AND document.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Because a delta can be a maximum of 2 bytes (depending on the packet status). Will document.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Will do.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bgrozev: No, that's for property names.

src/org/jitsi/impl/neomedia/transform/OriginalHeaderBlockTransformEngine.java
bgrozev: Because the first byte contains the ID/length fields, data follows. Will document.

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
bbaldino: oops, my edit below was supposed to be on this comment:

"EDIT: i see now that was another place, but maybe it'd be good here too"

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
bgrozev: Looking into it now, if we were to use the same instance and clear, we would have to do consumption (creating an RTCP packet) while holding the lock. So I'd rather leave it as it is.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -162,7 +183,13 @@ public RTCPTCCPacket(long senderSSRC, long sourceSSRC,
         int packetCount
             = 1 + RTPUtils.sequenceNumberDiff(last.getKey(), firstSeq);
 
+        // Temporary buffer to store the fixed fields (8 bytes) and the list of
+        // packet status chunks (see the format above). The buffer be longer
```
bbaldino: nit: think there's a typo here: "The buffer be longer than needed" -> "The buffer may be longer than needed" (i think?)

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -0,0 +1,365 @@
...
+    /**
+     * Incoming transport-wide sequence numbers mapped to the timestamp of their
+     * reception (in milliseconds since the epoch).
+     */
+    private RTCPTCCPacket.PacketMap incomingPackets;
+
+    /**
+     * Used to synchronize access to {@link #incomingPackets}.
+     */
+    private final Object incomingSeqsSyncRoot = new Object();
```
bbaldino: do you want to rename this since you renamed incomingSeqs -> incomingPackets?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -0,0 +1,365 @@
...
+
+    @Override
+    public String toString()
+    {
+        return "RTCP transport-cc feedback";
+    }
+
+    /**
+     * An ordered collection which maps sequence numbers to timestamps, the
+     * order is my the sequence number.
```
bbaldino: "my" -> "by" ?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -87,36 +91,296 @@ public static PacketMap getPackets(ByteArrayBuffer baf)
...
+        int packetsRemaining = packetStatusCount;
+        while (packetsRemaining > 0)
+        {
+            if (pscOff + 2 > off + len)
+            {
+                logger.warn(PARSE_ERROR + "reached the end while reading chunks");
+                return null;
+            }
+
+            int packetsInChunk = getPacketCount(buf, pscOff);
```
bgrozev: No, it only counts the number of packets in the current chunk.

bbaldino: ah! i had completely missed it was declared right there.

bbaldino:
lgtm, but not feeling super-confident i was able to properly review the meat of the logic here, more just little code style and questions.  do you want me to merge @bgrozev or have someone else take a pass?

bgrozev:
I'm currently testing it and adding the last bits of RTCP parsing. I'll let you know when I think it's ready to merge.

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
```diff
@@ -237,10 +237,20 @@ public void assemble(DataOutputStream dataoutputstream)
     {
         dataoutputstream.writeByte((byte) (0x80 /* version */ | FMT));
         dataoutputstream.writeByte((byte) RTPFB);
-        dataoutputstream.writeShort(2 + (fci.length / 4));
+        int len = 2 /* sender + source SSRCs */ + (fci.length / 4);
+        if (fci.length % 4 != 0)
+        {
+            len++;
```
bbaldino: does this only need to be incremented once in this case?  or does it need to be aligned to some 4 byte multiple?

bbaldino: i see below we pad to word-align, but what's the `len++` for?

src/org/jitsi/impl/neomedia/rtcp/RTCPPacketParserEx.java
```diff
@@ -227,16 +227,24 @@ private RTCPFBPacket parseRTCPFBPacket(
 
         int fmt = firstbyte & 0x1f;
         if (type == RTCPFBPacket.RTPFB && fmt == NACKPacket.FMT)
+        {
```
bbaldino: 🥇

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -109,10 +113,7 @@ public static PacketMap getPacketsFci(ByteArrayBuffer fciBuffer)
         int off = fciBuffer.getOffset();
         int len = fciBuffer.getLength();
 
-        // The FCI starts right after the "SSRC of media source" field.
-        // We need at least 8 bytes for the fixed fields + 2 bytes for one
-        // packet status chunk.
-        if (len < 10)
+        if (len < MIN_FCI_LENGTH)
```
bbaldino: 👍

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -539,21 +518,22 @@ public RTCPTCCPacket(long senderSSRC, long sourceSSRC,
             else if (seqDelta % 7 == 3)
```
bbaldino: stuff like this seems very confusing, but i assume it's just how the spec defines things?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -642,16 +626,35 @@ else if (tsDelta < 8191 && tsDelta > -8192)
...
+        int len = 2 /* sender + source SSRCs */ + (fci.length / 4);
+        if (fci.length % 4 != 0)
+        {
+            len++;
+        }
+        dataoutputstream.writeShort(len);
         dataoutputstream.writeInt((int) senderSSRC);
         dataoutputstream.writeInt((int) sourceSSRC);
         dataoutputstream.write(fci);
+        for (int i = fci.length; i % 4 != 0; i++)
```
bbaldino: worth writing a helper for?

test/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacketTest.java
```diff
@@ -0,0 +1,99 @@
+/*
```
bbaldino: you're a hero

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
bgrozev: Yeah, we pad to align to a word, and len is the length in words. We need the ++ because "fci.length/4" rounds down to the nearest integer.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: I don't know how to simplify this code. I can add some documentation if you think it's necessary.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Yeah, I really should...

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
bbaldino: ah! now i understand (didn't realize it was the length in words)

bbaldino: is it worth renaming it, maybe?  i assumed it was the length in bytes.  `lenInWords` ?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bbaldino: if there's some stuff we could copy-paste from the spec, i think that'd be nice.  i know it's not you, just the standard, but seeing code like `if (seqDelta % 7 == 3)` just feels so arbitrary and weird to read

src/org/jitsi/impl/neomedia/rtcp/NACKPacket.java
bgrozev: Sounds good, will rename it as part of extracting this logic.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Will do.

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
bgrozev: Added.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Added

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Added.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Added docs (but on closer look there turned out to be no bug)

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Fixed.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Fixed.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -87,36 +91,296 @@ public static PacketMap getPackets(ByteArrayBuffer baf)
...
         }
 
         byte[] buf = fciBuffer.getBuffer();
         int off = fciBuffer.getOffset();
         int len = fciBuffer.getLength();
 
+        // The FCI starts right after the "SSRC of media source" field.
+        // We need at least 8 bytes for the fixed fields + 2 bytes for one
+        // packet status chunk.
+        if (len < 10)
```
bgrozev: Fixed.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
bgrozev: Cleaned this up.

src/org/jitsi/impl/neomedia/rtcp/RTCPFBPacket.java
```diff
@@ -162,71 +162,48 @@ public static ByteArrayBuffer getFCI(ByteArrayBuffer baf)
...
-
-        //  SSRC of media source: 32 bits
-        RTCPFeedbackMessagePacket.writeSSRC(sourceSSRC, buf, off);
-        off += 4;
-
-        if (fci != null && fci.length != 0)
+        // Full length in bytes, including padding.
+        int len = this.calcLength();
+        dataoutputstream.writeShort(len / 4 - 1);
+        dataoutputstream.writeInt((int) senderSSRC);
```
bbaldino: interpreting ssrc as an int here shouldn't cause any problems i guess, right?

src/org/jitsi/impl/neomedia/rtcp/RTCPFBPacket.java
```diff
@@ -162,71 +162,48 @@ public static ByteArrayBuffer getFCI(ByteArrayBuffer baf)
...
+        dataoutputstream.writeInt((int) sourceSSRC);
+        dataoutputstream.write(fci);
+
+        // Pad with zeros. Since the above fields fill in exactly 3 words,
+        // the number of padding bytes will only depend on the length of the
+        // fci field.
+        for (int i = fci.length; i % 4 != 0; i++)
```
bbaldino: at one point i think we talked about a helper function for padding?

if not, i wonder if it's clearer to write this as something like:
```
int neededPadding = fci.length % 4;
for (int i = 0; i < neededPadding; ++i)
{
   ...
```
but maybe not.  it's not bad as is, my brain just hiccups on the `i % 4 != 0`

PR #280, opened 2017-03-09
Corrected documentation on MediaServiceImpl

jcchuks:
Corrected the explanation of the boolean fields; net.java.sip.communicator.service.media.DISABLE\_AUDIO\_SUPPORT  and net.java.sip.communicator.service.media.DISABLE\_VIDEO\_SUPPORT

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -2,7 +2,7 @@
 
 ##### net.java.sip.communicator.service.media.DISABLE\_AUDIO\_SUPPORT=*boolean*
 
-Indicates whether the detection of audio CaptureDevices is to be disabled. The default value is false i.e. the detection of the audio CaptureDevices is disabled.
+Indicates whether the detection of audio CaptureDevices is to be disabled. The default value is false i.e. the detection of the audio CaptureDevices is enabled.
 
 ##### net.java.sip.communicator.impl.neomedia.audiosystem.DISABLED=*boolean*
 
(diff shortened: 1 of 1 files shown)


jcchuks:
Thanks, I have signed the documentation. I am hoping to participate on the project under GSoC.

bgrozev:
Hey @jcchuks 

Thank you for the contribution! We may need a few days to confirm the CLA on our end (it requires human work), but we'll add you to the jenkins list so that it doesn't complain any more.

About the specific changes: looking at the code (MediaServiceImpl.java), the default value for these properties is "false", and since the properties have "DISABLE" in their name, the default behavior is to enable the detection of audio/video devices. So it looks to me like the original documentation is correct. Does that make sense, or am I missing something?

Jenkins, add to whitelist.

jcchuks:
Thanks @bgrozev , 

What you said makes sense since default value is "false". I have just corrected the sentence _"the detection of the audio CaptureDevices is disabled"_ to  _"the detection of the audio CaptureDevices is enabled"_ in my fork.

I feel it reads better as  "_**the detection of the audio CaptureDevices is enabled**_" in the docs than "_the audio CaptureDevices are detected._".  Same for video.  

If the emboldened statement is sound (please correct me if am wrong), then I would ask that this pull request be closed, then I can make another pull request with the corrected version. 

Thanks.

bgrozev:
LGTM

bgrozev:
Got the CLA. Thanks again for the contribution.

jcchuks:
My pleasure. Thanks.

PR #285, opened 2017-03-14
Reduce FIR spamming

gpolitis:


src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -127,7 +127,7 @@ public MediaStreamTrackReceiver getMediaStreamTrackReceiver()
      * @param frameDesc
      * @param nowMs
```
bbaldino: docs need updating...it'd be nice to have a good definition for what `isNew` really denotes...i'm not sure i know.  is it the first frame after a stream either 1) begins for the first time or 2) resumes after being suspended?

gpolitis:
@bbaldino could you please have a look and eventually this?

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -124,10 +124,14 @@ public MediaStreamTrackReceiver getMediaStreamTrackReceiver()
      * Updates rate statistics for the encodings of the tracks that this
      * receiver is managing. Detects simulcast stream suspension/resuming.
      *
-     * @param frameDesc
-     * @param nowMs
+     * @param pkt the received RTP packet that caused the update method call.
+     * @param frameDesc the {@link FrameDesc} to which the RTP packet belongs.
+     * @param frameIsNew true if the packet is from a frame that we have not
```
bbaldino: Ευχαριστώ πολύ!

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -150,6 +154,13 @@ void update(RawPacket pkt, FrameDesc frameDesc, long nowMs)
 
         if (!frameDesc.isIndependent())
         {
+            if (!frameIsNew)
+            {
+                // We check for stream suspension whenever we receive a new
```
bbaldino: nit: maybe this is better as `We only check for stream suspension if this is the first packet of a new frame`?

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -452,8 +452,10 @@ void update(RawPacket pkt, long nowMs)
         long ts = pkt.getTimestamp();
         FrameDesc frame = frames.get(ts);
 
+        boolean isNew;
```
bbaldino: could we rename to something like `isFirstPacketOfNewFrame` maybe?  also, if we initialize it to false we can get rid of the else block below

src/org/jitsi/impl/neomedia/rtp/translator/SimulcastController.java
```diff
@@ -274,17 +281,34 @@ public void setTargetIndex(int newTargetIdx)
...
             }
 
+            RTPEncodingDesc[] sourceEncodings = sourceTrack.getRTPEncodings();
+
             int currentTL0Idx = bitstreamController.getCurrentIndex();
             if (currentTL0Idx > -1)
             {
-                currentTL0Idx = sourceTrack.getRTPEncodings()
-                    [currentTL0Idx].getBaseLayer().getIndex();
+                currentTL0Idx
```
bbaldino: just curious, what's the meaning behind the "TL0" ?

src/org/jitsi/impl/neomedia/rtp/translator/SimulcastController.java
```diff
@@ -260,6 +260,13 @@ public boolean accept(byte[] buf, int off, int len)
      */
     public void setTargetIndex(int newTargetIdx)
     {
+        int oldTargetIdx = bitstreamController.getTargetIndex();
+        if (oldTargetIdx == newTargetIdx)
+        {
+            // Do not spam with FIRs.
```
bbaldino: this comment is a bit confusing.  i think i understand that if the new index is the same we don't need to do anything else, so maybe this comment should be something more broad like `No work necessary`?

src/org/jitsi/impl/neomedia/rtp/translator/SimulcastController.java
gpolitis: Temporal Layer 0; simulcast doesn't have to know anything about higher temporal layers.

src/org/jitsi/impl/neomedia/rtp/translator/SimulcastController.java
```diff
@@ -274,17 +281,34 @@ public void setTargetIndex(int newTargetIdx)
...
+                    = sourceEncodings[currentTL0Idx].getBaseLayer().getIndex();
             }
 
             int targetTL0Idx
-                = sourceTrack.getRTPEncodings()[newTargetIdx].getBaseLayer().getIndex();
+                = sourceEncodings[newTargetIdx].getBaseLayer().getIndex();
 
-            if (currentTL0Idx != targetTL0Idx)
+            // Something lower than the current must be streaming.
+            boolean sendFIR = targetTL0Idx < currentTL0Idx;
```
bbaldino: trying to make sure i'm understanding this code correctly:

1. If the target index is less than the current index, we can assume the target index stream must exist (since layers 'below' the current layer can be assumed to be sending as well
2. If the target index is greater than the current index, we need to check if target layer stream is actually sending. [if it isn't, we'll keep the target as the higher layer but fall back to the next one in other parts of the code, is that right?]

i wonder if it's just easier to do:
`sendIDR = sourceEncodings[i].isActive()` ?

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
gpolitis: I used `isPacketOfNewFrame` instead as it doesn't have to be the first one.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -452,8 +452,10 @@ void update(RawPacket pkt, long nowMs)
         long ts = pkt.getTimestamp();
         FrameDesc frame = frames.get(ts);
 
+        boolean isPacketOfNewFrame;
```
bbaldino: sorry i'm being nit-picky, but i think the 'first' part (`isFirstPacketOfNewFrame`) of the variable is important, right?

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
bbaldino: oops, didn't see your reply here when i made my follow-up comment.  ok sounds good!

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
bbaldino: ignore this, saw your reply above

src/org/jitsi/impl/neomedia/rtp/translator/SimulcastController.java
gpolitis: That's a good suggestion but if `targetTL0Idx` is inactive, we might still have a chance with some other `idx` that satisfies `currentTL0Idx < idx < targetTL0Idx`. For example, if currentTL0Idx refers to 180p@3.5fps and targetTL0Idx refers to 720p@30fps, and targetTL0Idx is inactive, but 360p@30fps is streaming, we still want to ask for a FIR.

bbaldino: ah i see.  ok

bbaldino:
lgtm once tests look good

PR #287, opened 2017-03-17
fix: Updates a field from a single thread to avoid potential sync issues

gpolitis:


src/org/jitsi/impl/neomedia/rtp/translator/SimulcastController.java
```diff
@@ -147,6 +147,20 @@ public long getTargetSSRC()
      */
     public boolean accept(byte[] buf, int off, int len)
     {
+        int targetIndex = bitstreamController.getTargetIndex()
+            , currentIndex = bitstreamController.getCurrentIndex();
+
+        if (targetIndex < 0 && currentIndex > -1)
+        {
+            synchronized (this)
```
bbaldino: i didn't see any `synchronized` go away in these changes?

gpolitis: We want to keep some synchronization when we switch from one bitstream controller to another so that the target and the optimal index are not messed up. What changed here is that we only write the this.bitstreamController field from the filter thread.

bbaldino: i'm not sure the synchronized here is helping much, is it?  it's possible `setTargetIndex` is called right before this (when the bitrateController is re-assigned).  from what i understand, we could do it this way and get rid of the need for synchronization:

```
BitstreamController new = new BitstreamController(bitstreamController, -1);
this.bitstreamController = new;
```
does that seem correct?

gpolitis: No, unfortunately that wouldn't work. Suppose we remove the synchronized. And suppose that the filter thread has finished creating a new BitstreamController and is about to assign it to the instance field. Now, suppose another thread is calling SimulcastController.setTargetIndex which will call the BitstreamController.setTargetIndex of the old bitstream controller. The filter thread is left with a stale targetIndex and we don't want that.

bbaldino: but how is that different from `setTargetIndex` being called right before we do this?  does this new assignment carry the targetIndex over from the previous object?

gpolitis: the `setTargetIndex` is also synchronized

bbaldino: sorry, i answered my own question and see that we're copying over the `targetIndex` value from the old one when we do `new BitstreamController(bitstreamController, -1);`.  i see now.

bbaldino: so why do we create a new BitstreamController instance in this scenario?

gpolitis: This is for the case where we want to suspend a stream. We handle stream suspension just like a simulcast switch to another RTP stream, so we get keyframe management for free (i.e. when a stream gets re-activated, it needs to start with a key frame). This is exactly the purpose of the SimulcastController and it would be a pity to re-implement this logic in the BitstreamController.

gpolitis: @bbaldino would you like to merge this, if it looks good?

gpolitis:
Replaced by https://github.com/jitsi/libjitsi/pull/288

src/org/jitsi/impl/neomedia/rtp/translator/SimulcastController.java
gpolitis: @bbaldino I followed your suggestion in https://github.com/jitsi/libjitsi/pull/288. Thanks!

PR #288, opened 2017-03-21
Adaptivity fixes

gpolitis:


src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -120,6 +125,35 @@ public FrameDesc put(Long key, FrameDesc value)
...
+        = new TreeMap<Long, FrameDesc>()
+    {
+        /**
+         * A helper {@link LinkedList} that is used to cleanup the map.
+         */
+        private LinkedList<Long> tsl = new LinkedList<>();
+
+        /**
+         * {@inheritDoc}
+         */
```
bbaldino: this must be doing something extra than the parent, right? some extra docs here describing what extra logic it has would be good

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -176,15 +210,20 @@ public RTPEncodingDesc(
...
      */
     public RTPEncodingDesc(
         MediaStreamTrackDesc track, int idx,
         long primarySSRC, long rtxSSRC,
         int tid, int sid,
+        int resolution,
         RTPEncodingDesc[] dependencyEncodings)
     {
+        // XXX we should be able to snif the actual resolution from the RTP
+        // packets.
```
bbaldino: how would we do that?

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -161,7 +195,7 @@ public RTPEncodingDesc(MediaStreamTrackDesc track, long primarySSRC)
     public RTPEncodingDesc(
         MediaStreamTrackDesc track, long primarySSRC, long rtxSSRC)
     {
-        this(track, 0, primarySSRC, rtxSSRC,
+        this(track, 0, primarySSRC, rtxSSRC, -1 /* resolution */,
```
bbaldino: if `-1` is the base/invalid resolution, i think a constant would be nice

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
gpolitis: We could do this by inspecting the payload (of keyframes). This wouldn't work with PERC tho, in which case we should get it from signaling.

gpolitis:
@bbaldino thanks very much for the review and all your suggestions 👍 Let's wait for @bgrozev to have a say.

gpolitis:
@bbaldino I talked to @bgrozev and he's done with the review. Could you please merge this so that I can update the JVB pom file?

PR #294, opened 2017-03-24
Fixes warnings and adds a new method used by the JVB.

gpolitis:


src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -121,6 +121,39 @@ public MediaStreamTrackReceiver getMediaStreamTrackReceiver()
     }
 
     /**
+     * Gets the optimal bitrate (bps). Together with the current bitrate, it
+     * allows to calculate the probing bitrate.
+     *
+     * @return the optimal bitrate (bps).
+     */
+    public long getBps(int idx)
```
bbaldino: why not call it `getOptimalBps`?  also, can you add a comment to define what the definition of 'optimal bitrate' is?

src/org/jitsi/impl/neomedia/transform/DebugTransformEngine.java
```diff
@@ -273,22 +273,16 @@ private RawPacket transform(RawPacket pkt, boolean data, boolean sender)
             {
                 len = pkt.getHeaderLength();
             }
-            else // if (RTCPPacketPredicate.INSTANCE.test(pkt))
+            // else if (RTCPPacketPredicate.INSTANCE.test(pkt))
```
bbaldino: do we need this block at all?

PR #295, opened 2017-03-24
feat: Adds a utility method.

bgrozev:


src/org/jitsi/util/ConfigUtils.java
```diff
@@ -228,4 +228,30 @@ public static String getString(
...
+     * @param cfg the {@code ConfigurationService} to get the value from or
+     * {@code null} if the property is to be retrieved from {@code System}
+     * @param property the name of the property to get
+     * @param propertyAlternative an alternative name of the property
+     * @param defaultValue the value to be returned if {@code property} is not
+     * associated with a value
+     * @return the value as a {@code String} of {@code property} retrieved from
+     * either {@code cfg} or {@code System}
+     */
+    public static String getString(ConfigurationService cfg,
```
bbaldino: we have properties with alternative names?

bgrozev: This is perhaps not phrased in the best way. The way I use it is with property names that were renamed, but the old names were kept as deprecated (in order to not break people's existing configurations).

bbaldino: 👍

PR #297, opened 2017-04-07
Bandwidth estimation and probing for bandwidth fixes

gpolitis:
This PR is required for https://github.com/jitsi/jitsi-videobridge/pull/446. For context on why this PR is necessary and what this PR tries to achieve, have a look at the JVB PR description.

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
+++ b/src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
@@ -31,8 +31,11 @@ import org.jitsi.util.*;
 public class MediaStreamTrackDesc
 {
     /**
-     * The system property name that holds the minimum time (in millis) that is
-     * required for the media engine to generate a new key frame.
+     * The system property name that holds the maximum frequency (in millis) at
+     * which the media engine generates key frame.
+     *
+     * We take that into account to identify batches of key frames which helps
+     * us identify which simulcast flows are active/inactive.
      */
     public static final String MIN_KEY_FRAME_WAIT_MS_PNAME
         = "org.jitsi.impl.neomedia.rtp.MediaStreamTrackDesc" +
@@ -45,6 +48,18 @@ public class MediaStreamTrackDesc
     public static final String SUSPENSION_THRESHOLD_MS_PNAME
         = "org.jitsi.impl.neomedia.rtp.MediaStreamTrackDesc" +
             ".SUSPENSION_THRESHOLD_MS";
+
+    /**
+     * The default maximum frequency (in millis) at which the media engine
+     * generates key frame.
+     */
+    private static final int MIN_KEY_FRAME_WAIT_MS_DEFAULT = 300;
+
+    /**
+     * The default maximum time interval (in millis) an encoding can be
+     * considered active without new frames.
+     */
+    private static final int SUSPENSION_THRESHOLD_MS_DEFAULT = 600;
 
     /**
      * The {@link Logger} used by the {@link MediaStreamTrackDesc} class and its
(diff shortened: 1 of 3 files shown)


src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -29,23 +31,47 @@
 public class MediaStreamTrackDesc
 {
     /**
+     * The system property name that holds the minimum time (in millis) that is
+     * required for the media engine to generate a new key frame.
```
bbaldino: `generate a new key frame`? does that mean request a key frame?

bbaldino: also, i assume the 'wait' here is from a previous keyframe request?  or is this a delay from when the user joins/stream is created?

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -29,23 +31,47 @@
...
+    private static final ConfigurationService
+        cfg = LibJitsi.getConfigurationService();
+
+    /**
      * The minimum time (in millis) that is required for the media engine to
      * generate a new key frame.
      */
-    private static final int MIN_KEY_FRAME_WAIT_MS = 300;
+    private static final int MIN_KEY_FRAME_WAIT_MS
+        = cfg != null ? cfg.getInt(MIN_KEY_FRAME_WAIT_MS_PNAME, 300) : 300;
```
bbaldino: why not do a static var for the default value as well?

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -29,23 +31,47 @@
...
 
     /**
      * The maximum time interval (in millis) an encoding can be considered
      * active without new frames.
      */
-    private static final int SUSPENSION_THRESHOLD_MS = 600;
+    public static final int SUSPENSION_THRESHOLD_MS
+        = cfg != null ? cfg.getInt(SUSPENSION_THRESHOLD_MS_PNAME, 600) : 600;
```
bbaldino: same as above

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -234,8 +240,8 @@ public RTPEncodingDesc(
      */
     private static void applyFrameBoundsHeuristics(FrameDesc a, FrameDesc b)
     {
-        int end = a.getEnd(), start = b.getStart();
-        if (end != -1 && start != -1)
+        int aEnd = a.getEnd(), bStart = b.getStart();
```
bbaldino: do these represent timestamps?

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -347,6 +358,37 @@ public boolean isActive()
...
+    {
+        if (active && performTimeoutCheck)
+        {
+            if (lastReceivedFrame == null)
+            {
+                return false;
+            }
+            else
+            {
+                long silentIntervalMs
```
bbaldino: 'silent' here is referring to no video packets received?  might be a bit confusing with 'audio silence'.  `timeSinceLastReceivedFrameMs` maybe?

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
gpolitis: The max frequency that the webrtc engine generates a key frame is 300 ms. We take that into account to identify batches of key frames which helps us identify which simulcast flows are active/inactive.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
gpolitis: Those are sequence numbers.

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
gpolitis: MAX_KEY_FRAME_FREQ_MS?

bbaldino: ah i see.  so the client-side stack won't generate them more frequently than that, so we don't bother requesting.  i think it's worth documenting that, especially since that's technically chrome-specific, right?  i think the variable name is ok as is, just updating the comment a bit to clarify.  i was just confused by 'generate' since we don't generate key frames here, just request them.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
bbaldino: gotcha...i see this was existing, but having the unit in there might help? i.e. `aEnd` -> `aLastSeqNum` or something.  just a thought

PR #298, opened 2017-04-17
fix: Fixes the length of the abs-send-time extension

bgrozev:
added to packets which don't already have an extension.

### DIFF
--- a/src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
+++ b/src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
@@ -68,7 +68,7 @@ public class AbsSendTimeEngine
                 = pkt.getHeaderExtension((byte) extensionID);
             if (ext == null)
             {
-                ext = pkt.addExtension((byte) extensionID, 2);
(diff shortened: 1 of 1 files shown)


src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
```diff
@@ -68,7 +68,7 @@ public RawPacket transform(RawPacket pkt)
                 = pkt.getHeaderExtension((byte) extensionID);
             if (ext == null)
             {
-                ext = pkt.addExtension((byte) extensionID, 2);
+                ext = pkt.addExtension((byte) extensionID, 3);
```
bbaldino: worth adding a constant for what this number represents?

bgrozev: Added

bbaldino:
👍 will merge when tests are done

bgrozev:
@gp, thank you for reporting this! Do you need this included in jitsi-videobridge master right away?

gpolitis:
@bgrozev not right away, but it would be nice to have it for tomorrows standup

PR #299, opened 2017-04-20
Fix framemarking

bgrozev:


saghul:
🚔 🚔 🚔  Fix what exactly? :-)

bgrozev:
Uses framemarking, if it is available, to determine start and end of frames (as opposed to inspecting the payload (and M-bit)). Noted that I should include this in the commit/PR message.

murillo128:
https://tools.ietf.org/html/draft-ietf-avtext-framemarking-04

This is helpful for SFUs to avoid parsing the payload and a must-have if payload is encrypted as in PERC

Framemarking was partially implemented as only IFRAME was checked using the header extensions but start/end check was still done on payload

PR #301, opened 2017-04-25
feat: Adds methods for VP8 PLD fields manipulation.

gpolitis:


src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
```diff
@@ -542,20 +542,91 @@ public static int getSize(byte[] input, int offset)
...
+         * @param off the offset in the byte buffer where the VP8 payload
+         * starts.
+         * @param len the length of the VP8 payload in the byte buffer.
+         * @return true if the VP8 payload contains a KEYIDX, false
+         * otherwise.
+         */
+        public static boolean hasKEYIDX(byte[] buf, int off, int len)
+        {
+            return (buf[off] & X_BIT) != 0
+                && ((buf[off + 1] & T_BIT) != 0 || (buf[off + 1] & K_BIT) != 0);
```
bgrozev: According to the spec "If T (above) is set to 1 but K is set to 0, the TID/Y/KEYIDX octet MUST be present, but the KEYIDX field MUST be ignored."

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
```diff
@@ -542,20 +542,91 @@ public static int getSize(byte[] input, int offset)
...
+         * otherwise.
+         */
+        public static boolean hasPictureId(byte[] buf, int off, int len)
+        {
+            if (!isValid(buf, off))
+                return false;
+
+            if ((buf[off] & X_BIT) == 0
+                || (buf[off+1] & I_BIT) == 0)
+                return false;
```
bgrozev: I would simplify to "return isValid(buf,off) && (buf[off] & X_BIT != 0) && (buf[off+1] & I_BIT) != 0)"

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
```diff
@@ -542,20 +542,91 @@ public static int getSize(byte[] input, int offset)
...
+         * @param buf the byte buffer that contains the VP8 payload.
+         * @param off the offset in the byte buffer where the VP8 payload
+         * starts.
+         * @param len the length of the VP8 payload in the byte buffer.
+         * @return true if the VP8 payload contains a TL0PICIDX, false
+         * otherwise.
+         */
+        public static boolean hasTL0PICIDX(byte[] buf, int off, int len)
+        {
+            return (buf[off] & X_BIT) != 0 && (buf[off+1] & L_BIT) != 0;
```
bgrozev: Why is the isValid() check missing only from this method?

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
```diff
@@ -566,6 +637,60 @@ private static int getPictureId(byte[] input, int offset)
...
+            }
+
+            buf[off + 2] = (byte) (0x80 | (val >> 8) & 0x7F);
+            buf[off + 3] = (byte) (val & 0xFF);
+
+            return true;
+        }
+
+        /**
+         * Sets the TL0PICIDX field for the VP8 payload specified in the
```
bgrozev: Nit pick: its a VP8 Payload Descriptor, not the payload itself.

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
```diff
@@ -566,6 +637,60 @@ private static int getPictureId(byte[] input, int offset)
...
+            }
+
+            int sz = getSize(buf, off);
+            if (hasKEYIDX(buf, off, len))
+            {
+                buf[off + sz - 2] = (byte) val;
+            }
+            else
+            {
+                buf[off + sz - 1] = (byte) val;
```
bgrozev: In light of the comment above about KEYIDX, I would change this find the offset based on the presence (and length) of PictureID

bgrozev:
Something related, although not introduced with this PR: getTemporalLayerIndex() returns  "(buf[off + sz - 1] & 0xc0) >> 6". According to RFC7741 the TID field is 3 bits long, so this should be "&0xe0 >> 5"

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
gpolitis: isValid is a dummy method, it doesn't do anything unfortunately :(

bgrozev:
> Something related, although not introduced with this PR: getTemporalLayerIndex() returns "(buf[off + sz - 1] & 0xc0) >> 6". According to RFC7741 the TID field is 3 bits long, so this should be "&0xe0 >> 5

Disregard this, I misread the spec.

PR #308, opened 2017-05-18
feat: Adds a frame rate field to the RTP encoding.

gpolitis:


src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -83,11 +88,18 @@
...
+     * The max height of the bitstream that this instance represents. The actual
+     * height may be less due to bad network or system load.
      */
     private final int height;
 
     /**
+     * The max frame rate of the bitstream that this instance represents. The
+     * actual frame rate may be less due to bad network or system load.
+     */
+    private final double frameRate;
```
bgrozev: Can you please add a note on what the units are (frames per second?)?

Issue #318, opened 2017-06-09
Debug logging can cause init to fail.


guusdk:
Although wildly unexpected, I frequently have jicofo that fails to start up, as the property iterator in `org.jitsi.impl.configuration.ConfigurationServiceImpl#debugPrintSystemProperties` throws a ConcurrentModificationException.

Obviously, the root cause of the problem is another thread modifying settings, which should be solved too. But, regardless, a failure to write debug information should not cause the initialization to be aborted.

```
INFO: Failed to initialize service implementation org.jitsi.impl.configuration.ConfigurationServiceImpl.
java.util.ConcurrentModificationException
	at java.util.Hashtable$Enumerator.next(Hashtable.java:1378)
	at org.jitsi.impl.configuration.ConfigurationServiceImpl.debugPrintSystemProperties(ConfigurationServiceImpl.java:1399)
	at org.jitsi.impl.configuration.ConfigurationServiceImpl.<init>(ConfigurationServiceImpl.java:156)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at org.jitsi.impl.libjitsi.LibJitsiImpl$ServiceLock.initializeService(LibJitsiImpl.java:196)
	at org.jitsi.impl.libjitsi.LibJitsiImpl$ServiceLock.getService(LibJitsiImpl.java:131)
	at org.jitsi.impl.libjitsi.LibJitsiImpl.getService(LibJitsiImpl.java:91)
	at org.jitsi.impl.libjitsi.LibJitsiOSGiImpl.getService(LibJitsiOSGiImpl.java:95)
	at org.jitsi.service.libjitsi.LibJitsi.invokeGetServiceOnImpl(LibJitsi.java:172)
	at org.jitsi.service.libjitsi.LibJitsi.getConfigurationService(LibJitsi.java:94)
	at net.java.sip.communicator.impl.configuration.ConfigurationActivator.start(ConfigurationActivator.java:99)
	at org.jitsi.impl.osgi.framework.BundleImpl.start(BundleImpl.java:293)
[...]

PR #319, opened 2017-06-09
318: Problems generating debug info should not stop initialization.

guusdk:
When an exception is thrown while generating debug information, whatever started the process should not stop, as illustrated in https://github.com/jitsi/libjitsi/issues/318

This commit catches all unchecked exceptions that occur while generating debug information.


guusdk:
If this gets merged, please update the jicofo dependency to libjitsi to include this change.

PR #320, opened 2017-06-12
feat: Disable RTCP termination for 1:1 calls.

gpolitis:


bbaldino:
what's the win like for doing this?

gpolitis:
The sender adapts to the end-to-end link, and not the sender-bridge link.

bbaldino:
seems like that wouldn't be very valuable unless we disabled simulcast, too--are we doing that?

gpolitis:
That would be the right thing to do, but we cannot easily disable simulcast and then re-enable it because of this Chrome bug https://bugs.chromium.org/p/webrtc/issues/detail?id=7555.

bbaldino:
but if we can't disable simulcast, does disabling rtcp termination help much?

gpolitis:
why wouldn't it help? if the receiver is sitting on a poor link, what's the benefit of the sender sending at high bitrate?

bbaldino:
but won't it already just use the existing logic to forward a lower simulcast layer?  i guess this allows for more granular bitrate changes than jumping layers, i was just wondering if the new code paths introduced were worth the tradeoff.

gpolitis:
I lost you. Could you please help me understand your question?

gpolitis:
Actually this won't work at all because of adaptivity and bandwidth probing.

PR #326, opened 2017-06-16
Fix FrameMarking isKeyframe

murillo128:
Bit operation was wrong, all frames were marked as I frames (S bit set) 

cc/ @bgrozev


bgrozev:
Good catch, thanks!

murillo128:
BTW, I am preparing another PR for using framemarking for retrieving VP8 getTemporalId which is still looking at the media data

PR #327, opened 2017-06-16
Fix RED packets with a single block length

murillo128:
When the RED packet contains a single block, we just get the PT and remove the first byte from the buffer.

The length was being calculated wrongly and it was negative. The check is also removed as it is redundant as if buffer.length > off + len then buffer.length  > (off+1) + (len -1)

cc/ @bgrozev


bgrozev:
Jenkins add to whitelist

bgrozev:
LGTM, @gpolitis can you also take a look? I wonder if we should also check for length >= 1 before reading the blockPT byte.

PR #332, opened 2017-06-21
Retrieve Spatial and Temporal layer ids from frame marking header extension

murillo128:
The patch was going to be more ambitious as IMHO the RawPacket should be passed from the top layers instead of buff,off,len ... but it was not working so I decided to go KISS first

In order to have simulcast enabled with media encryption it is required to disable the vp8 extended pic id on videobridge:
```
org.jitsi.videobridge.ENABLE_VP8_PICID_REWRITING=false
```

cc/ @bgrozev

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -3787,12 +3787,6 @@ public class MediaStreamImpl
             {
                 return FrameMarkingHeaderExtension.isStartOfFrame(fmhe);
             }
-
-            if (logger.isDebugEnabled())
-            {
-                logger.debug("Packet with no frame marking, while frame marking"
-                             + " is enabled.");
-            }
             // Note that we go on and try to use the payload itself. We may want
             // to change this behaviour in the future, because it will give
             // wrong results if the payload is encrypted.
@@ -3857,12 +3851,6 @@ public class MediaStreamImpl
             if (fmhe != null)
             {
                 return FrameMarkingHeaderExtension.isEndOfFrame(fmhe);
-            }
-
-            if (logger.isDebugEnabled())
-            {
-                logger.debug("Packet with no frame marking, while frame marking"
-                             + " is enabled.");
             }
             // Note that we go on and try to use the payload itself. We may want
             // to change this behaviour in the future, because it will give
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -3653,6 +3653,21 @@ public class MediaStreamImpl
      */
     public int getTemporalID(byte[] buf, int off, int len)
     {
+        if (frameMarkingsExtensionId != -1)
+        {
+            RawPacket pkt = new RawPacket(buf, off, len);
+            RawPacket.HeaderExtension fmhe
+                = pkt.getHeaderExtension(frameMarkingsExtensionId);
+
+            if (fmhe != null)
+            {
+                return FrameMarkingHeaderExtension.getTemporalID(fmhe);
+            }
+            // Note that we go on and try to use the payload itself. We may want
+            // to change this behaviour in the future, because it will give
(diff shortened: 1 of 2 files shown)


src/org/jitsi/impl/neomedia/rtp/FrameMarkingHeaderExtension.java
```diff
@@ -105,4 +119,26 @@ public static boolean isEndOfFrame(ByteArrayBuffer baf) {
...
+        {
+            return 0;
+        }
+        /*
+         * THIS IS STILL NOT YET CORRECTLY DEFINED ON FRAMEMARKING DRAFT!
+         */
+        return 0;
+    }
+
+    public static byte getTemporalID(ByteArrayBuffer baf) {
```
bgrozev: We're gonna need javadocs on the methods.

src/org/jitsi/impl/neomedia/rtp/FrameMarkingHeaderExtension.java
```diff
@@ -105,4 +119,26 @@ public static boolean isEndOfFrame(ByteArrayBuffer baf) {
...
+        if (baf == null || baf.getLength() < 2)
```
bgrozev: According to the draft (avtext, -4) the TID is also only present in the scalable version. Should we return -1 as an indication that the field is not present if length<4?

src/org/jitsi/impl/neomedia/rtp/FrameMarkingHeaderExtension.java
```diff
@@ -105,4 +119,26 @@ public static boolean isEndOfFrame(ByteArrayBuffer baf) {
         byte b = baf.getBuffer()[baf.getOffset() + 1];
         return (b & E_BIT) != 0;
     }
+
+    public static byte getSpatialID(ByteArrayBuffer baf, String encoding) {
+        // Only present on scalable version
+        if (baf == null || baf.getLength() < 4)
+        {
+            return 0;
```
bgrozev: Should we return -1 instead? I thought that 0 was a valid ID.

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -3653,6 +3653,21 @@ public void injectPacket(RawPacket pkt, boolean data, TransformEngine after)
      */
     public int getTemporalID(byte[] buf, int off, int len)
     {
+        if (frameMarkingsExtensionId != -1)
+        {
+            RawPacket pkt = new RawPacket(buf, off, len);
```
bgrozev: I would like to get this refactored eventually, but I guess it is OK for now.

bgrozev:
> The patch was going to be more ambitious as IMHO the RawPacket should be passed from the top layers instead of buff,off,len

I just discussed this with George and we agree. We can merge this PR as it is, but contributions that work towards passing the RawPacket instance are also welcome.

src/org/jitsi/impl/neomedia/rtp/FrameMarkingHeaderExtension.java
murillo128: If the field is not present then the layer Id is 0 according to the draft.

src/org/jitsi/impl/neomedia/rtp/FrameMarkingHeaderExtension.java
murillo128: Yes, but as before, if field is not present, layer is 0 according to the draft

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
murillo128: Me too, I started with the patch, but if I created the RawPacket outside of the function and pass it to the inner layer, everything stop working. I was getting crazy, so I decided to keep it for later.. :)

murillo128:
Is it possible to merge this PR as it is? or you would require any modification to it?

bgrozev:
Just the missing javadocs (it's a policy), everything else looks good.

murillo128:
done.

PR #334, opened 2017-06-26
Remote_Bitrate_Estimator_Abs_send_time

jcchuks:


src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
```diff
@@ -240,6 +260,14 @@ public void update(
...
         prevOffset = offset;
         offset = offset + K[1] * residual;
+
+        /**
+         * Migrated from webrtc. Used for logging
+         */
+//        BWE_TEST_LOGGING_PLOT(1, "kc", nowMs, K[0]);
+//        BWE_TEST_LOGGING_PLOT(1, "km", nowMs, K[1]);
+//        BWE_TEST_LOGGING_PLOT(1, "slope_1/bps", nowMs, slope);
+//        BWE_TEST_LOGGING_PLOT(1, "var_noise", nowMs, varNoise_);
```
bgrozev: I don't think we need any of the changes to this file -- all usage of nowMs is commented out.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
+package org.jitsi.impl.neomedia.rtp.remotebitrateestimator;
```
bgrozev: Can you please add the license header (see any of the other .java files in the project).

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+
+import org.ice4j.util.RateStatistics;
+import org.jitsi.impl.neomedia.codec.audio.silk.K2aFLP;
+import org.jitsi.impl.neomedia.transform.PacketTransformer;
+import org.jitsi.service.neomedia.RawPacket;
+import org.jitsi.service.neomedia.rtp.RemoteBitrateEstimator;
+import org.jitsi.util.Logger;
+
+import javax.media.rtp.RTPHeader;
+import java.util.*;
```
bgrozev: Can you please remove the unused imports and convert them to package imports (e.g. org.ice4j.util.*)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+
+/**
+ * webrtc.org abs_send_time implementation as of June 26, 2017.
```
bgrozev: If you know the commit ID in webrtc.org it would be nice to add it here.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+ * @author Julian Chukwu
+ */
+public class RemoteBitrateEstimatorAbsSendTime
+    implements RemoteBitrateEstimator
+{
+    private static final Logger logger = Logger
+            .getLogger(RemoteBitrateEstimatorAbsSendTime.class);
+
+    private enum probeResult
```
bgrozev: The convention is to name classes and enums with the first letter capitalized (ProbeResult)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+
+    public final static int kTimestampGroupLengthMs = 5;
+    public final static int kAbsSendTimeFraction = 18;
+    public final static int kAbsSendTimeInterArrivalUpshift = 8;
+    public final static int kInterArrivalShift
+            =  kAbsSendTimeFraction + kAbsSendTimeInterArrivalUpshift;
+    public final static int kInitialProbingIntervalMs = 2000;
+    public final static int kMinClusterSize = 4;
+    public final static int kMaxProbePackets = 15;
+    public final static int kExpectedNumberOfProbes = 3;
```
bgrozev: Are there going to be used from outside this class? If not, they should be private.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+
+    private static final double kTimestampToMs = 1000.0 /
+            (1 << kInterArrivalShift) ;
+
+    public <K,V> List<K> Keys(TreeMap<K,V> _map)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    {
+        ArrayList<K> keys = new ArrayList<K>();
+        for(Map.Entry<K,V> entry : _map.entrySet())
+        {
+            keys.add(entry.getKey());
+        }
+        return keys;
+    }
+
+    long ConvertMsTo24Bits(long timeMs)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+
+    /**
+      * Confirm this on line 133 of remote_bitrate_estimator_abs_send_time.h
+     */
+    public TreeMap<Long,Long> ssrcs_ = new TreeMap<Long, Long>();
+    public static ArrayList<Probe> probes_ = new ArrayList<>();
+    public static long totalProbesReceived;
+    public static  long firstPacketTimeMs;
+    public static long lastUpdateMs;
+    public static boolean umaRecorded;
```
bgrozev: Everything should be private unless it specifically needs to be exposed (including the methods below)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+
+    boolean IsWithinClusterBounds(long sendDeltaMs, Cluster clusterAggregate)
+    {
+        if(clusterAggregate.count == 0)
+        {
+            return true;
+        }
+        double clusterMean = clusterAggregate.sendMeanMs /
+                (double)clusterAggregate.count;
+        /**
```
bgrozev: No need for the TODO

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    public static boolean umaRecorded;
+
+    /**
+     * @Todo confirm these
+     * clock and observer_ are defined as const
+     * e.g const Clock* const clock_;
+     * find out if there is need to make it final
+     */
+    public static long clock;
+    public static RemoteBitrateObserver observer_;
```
bgrozev: In java "static" defines a class field instead of an instance field. These should not be static. Try to make them final, if something prevents it leave it out.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
bgrozev: These should probably also not be "static"

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+     * /external/webrtc/+/master/webrtc/modules/remote_bitrate_estimator
+     * /remote_bitrate_estimator_abs_send_time.cc
+     */
+
+    void RemoteBitrateEstimatorAbsSendTime(RemoteBitrateObserver observer,
+                                           long time)
+    {
+        this.observer_ = observer;
+        this.clock = time;
+        this.interArrival = new InterArrival(90 * kTimestampGroupLengthMs,
```
bgrozev: Where do the arguments come from?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    public static long lastUpdateMs;
+    public static boolean umaRecorded;
+
+    /**
+     * @Todo confirm these
+     * clock and observer_ are defined as const
+     * e.g const Clock* const clock_;
+     * find out if there is need to make it final
+     */
+    public static long clock;
```
bgrozev: How can we replace the clock with a single "long" instance? We should use System.currentTimeMillis() instead of "clock", and remove the "clock" field.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
bgrozev: I don't think we need the "umaRecorded" field.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+        this.estimator = new OveruseEstimator(new OverUseDetectorOptions());
+        this.detector = new OveruseDetector(new OverUseDetectorOptions());
+        this.incomingBitrate = new RateStatistics(kBitrateWindowMs,8000);
+        this.incomingBitrateInitialized = false;
+        this.totalProbesReceived = 0;
+        this.firstPacketTimeMs = -1;
+        this.lastUpdateMs = -1;
+        this.umaRecorded = false;
+       // RTC_DCHECK(observer_);
+        logger.warn("RemoteBitrateEstimatorAbsSendTime: Instantiating.");
```
bgrozev: Can we make that an "info" log instead?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+        //network_thread_.DetachFromThread();
+    }
+
+    List<Cluster> computeClusters(List<Cluster> clusters)
```
bgrozev: Why was it necessary to change this to return a list?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+                    bestIt = cluster;
+                }
+            }
+            else
+            {
+                /**
+                 * @Todo
+                 * What do we loose by converting from double to long ?
+                 */
+                double sendBitrateBps = cluster.meanSize * 8 * 1000
```
bgrozev: Looks like the worst that could happen is that the values in the log message are not precise

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    {
+        synchronized (critSect) {
+            List<Cluster> clusters = new ArrayList<Cluster>();
+            clusters = computeClusters(clusters);
+            if (clusters.isEmpty()) {
+                // If we reach the max number of probe packets and still
+                // have no clusters,
+                // we will remove the oldest one.
+                if (probes_.size() >= kMaxProbePackets)
+                    probes_.get(0);
```
bgrozev: get() will not remove an element from the list

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+                return probeResult.kNoUpdate;
+            }
+
+            Cluster bestProbe = findBestProbe(clusters);
+            if (bestProbe != null) {
```
bgrozev: findBestProbe will never return null

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+                        + " ms, num probes: " + cluster.count);
+                break;
+            }
+        }
+        return bestIt;
+    }
+
+    probeResult processClusters(long nowMs)
+    {
+        synchronized (critSect) {
```
bgrozev: Why do we synchronize here? The original code doesn't seem to do this

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+            // or finished with current set  of probes.
+            if (clusters.size() >= kExpectedNumberOfProbes)
+                probes_.clear();
+            return probeResult.kNoUpdate;
+        }
+    }
+
+    boolean isBitrateImproving(int newBitrateBps)
+    {
+        synchronized (critSect)
```
bgrozev: Why do we synchronize here? The original code doesn't seem to do this

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    {
+        //RTC_DCHECK(network_thread_.CalledOnValidThread());
+    /*if (!header.getExtension().hasAbsoluteSendTime) {
+        logger.warn( "RemoteBitrateEstimatorAbsSendTimeImpl: Incoming packet "
+        + "is missing absolute send time extension!";
+        return;
+    }
+    incomingPacketInfo(arrival_time_ms, header.getExtension().absoluteSendTime,
+            payload_size, header.ssrc);*/
+    }
```
bgrozev: We can leave this out for now

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+        long arrivalTimeMs,
+        long sendTime24bits,
+        long payloadSize,
+        long ssrc) {
+        //RTC_CHECK(sendTime24bits < (1L << 24));
+        if (!umaRecorded) {
+           // RTC_HISTOGRAM_ENUMERATION(kBweTypeHistogram,
+            // bweNames.kReceiverAbsSendTime, bweNames.kBweNamesMax);
+            umaRecorded = true;
+        }
```
bgrozev: This is unused

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    public final static int kAbsSendTimeInterArrivalUpshift = 8;
+    public final static int kInterArrivalShift
+            =  kAbsSendTimeFraction + kAbsSendTimeInterArrivalUpshift;
+    public final static int kInitialProbingIntervalMs = 2000;
+    public final static int kMinClusterSize = 4;
+    public final static int kMaxProbePackets = 15;
+    public final static int kExpectedNumberOfProbes = 3;
+
+    private static final double kTimestampToMs = 1000.0 /
+            (1 << kInterArrivalShift) ;
```
bgrozev: I think we may need to promote one of the arguments of the division to a double, e.g. 1000.0D

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+        long timestamp = sendTime24bits << kAbsSendTimeInterArrivalUpshift;
+        long sendTimeMs = (long) (timestamp * kTimestampToMs);
+        long nowMs = System.currentTimeMillis();
+        // TODO(holmer): SSRCs are only needed for REMB,
+        // should be broken out from  here.
+        // Check if incoming bitrate estimate is valid, and if it
+        // needs to be reset.
+        Long incomingBitrate_ =
+                incomingBitrate.getRate(arrivalTimeMs);
+        if (incomingBitrate_ != null)
```
bgrozev: getRate returns a "long" (small "l"), and converting this to a "Long" (big "L") will always be non-null. Use a "long" and check for != 0 (or > 0 if appropriate).

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+        incomingBitrate.update((int) payloadSize, arrivalTimeMs);
+        if (firstPacketTimeMs == -1) {
+            firstPacketTimeMs = nowMs;
+        }
+        long tsDelta = 0;
+        long tDelta = 0;
+        int sizeDelta = 0;
+        boolean updateEstimate = false;
+        long targetBitrateBps = 0;
+        Collection<Long> ssrcs = new ArrayList<Long>();
```
bgrozev: Lets remove the SSRCs, they are not used anywhere anyway

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+                updateEstimate = remoteRate.isValidEstimate();
+                ssrcs = Keys(ssrcs_);
+            }
+        }
+        if (updateEstimate) {
+            lastUpdateMs = nowMs;
+            /**
+             * @Todo Fix Problem with transforming Long to Integer
+             */
+            observer_.onReceiveBitrateChanged(null, targetBitrateBps);
```
bgrozev: This TODO is fixed now

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+        }
+    }
+
+    void Process() {}
+
+    long timeUntilNextProcess()
+    {
+        long kDisabledModuleTime = 1000;
+        return kDisabledModuleTime;
+    }
```
bgrozev: We don't need these unless we implement RecurringRunnable

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+
+    void timeoutStreams(long nowMs)
+    {
+        synchronized (critSect) {
```
bgrozev: Why do we synchronize here?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    void removeStream(long ssrc)
+    {
+        synchronized (critSect)
+        {
+            ssrcs_.remove(ssrc);
+        }
+    }
+
+    boolean latestEstimate(List<Long> ssrcs,
+        long bitrateBps) {
```
bgrozev: In the original bitrate_bps is a pointer. I don't know how it should be handled here (since it is currently not used)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    /**
+     * Returns the estimated payload bitrate in bits per second if a valid
+     * estimate exists; otherwise, <tt>-1</tt>.
+     *
+     * @return the estimated payload bitrate in bits per seconds if a valid
+     * estimate exists; otherwise, <tt>-1</tt>
+     */
+    @Override
+    public long getLatestEstimate() {
+        return 0;
```
bgrozev: Shouldn't we return the actual estimate?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+
+    /**
+     * Removes all data for <tt>ssrc</tt>.
+     *
+     * @param ssrc
+     */
+    @Override
+    public void removeStream(int ssrc)
+    {
+
```
bgrozev: Is this a wrapper for removeStream(long)?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,627 @@
...
+    }
+
+    public class Probe
```
bgrozev: Can you please move this below (after all the method definitions)

jitsi-developers:
On 28 June 2017 at 00:38, bgrozev <notifications@github.com> wrote:
>
> @bgrozev commented on this pull request.
>
> ________________________________
>
> In src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java:>
>
> >> +    }
> >>+
> >>+    private final Object critSect = new Object();
> >>+
> >>+    public final static int kTimestampGroupLengthMs = 5;
> >>+    public final static int kAbsSendTimeFraction = 18;
> >>+    public final static int kAbsSendTimeInterArrivalUpshift = 8;
> >>+    public final static int kInterArrivalShift
> >>+            =  kAbsSendTimeFraction + kAbsSendTimeInterArrivalUpshift;
> >>+    public final static int kInitialProbingIntervalMs = 2000;
> >>+    public final static int kMinClusterSize = 4;
> >>+    public final static int kMaxProbePackets = 15;
> >>+    public final static int kExpectedNumberOfProbes = 3;
> >>+
> >>+    private static final double kTimestampToMs = 1000.0 /
> >>+            (1 << kInterArrivalShift) ;
>
> >I think we may need to promote one of the arguments of the division to a double, e.g. 1000.0D

This is wrong.  1000.0 is already of type double.
http://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.10.2
says "A floating-point literal is of type float if it is suffixed with
an ASCII letter F or f; otherwise its type is double and it can
optionally be suffixed with an ASCII letter D or d" .

-- 
Martin

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Alright, makes sense

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: `this.clock` will be removed. `observer` is passed as a parameter when instantiating RBE_abs_send_time (as in webrtc).
the `InterArrival` parameters are default values. The same parameters were used in the RemoteBitrateEstimatorSingleStream. Also definition on line 34 of [Webrtc](https://chromium.googlesource.com/external/webrtc/+/master/webrtc/modules/remote_bitrate_estimator/inter_arrival.h )

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: I think method declaration of processClusters (in line 109 of [webrtc abs_send_time.h]
 (https://chromium.googlesource.com/external/webrtc/+/master/webrtc/modules/remote_bitrate_estimator/remote_bitrate_estimator_abs_send_time.h))uses locks.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Yes. Updated

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Am using `synchronized` as the equivalent of `EXCLUSIVE_LOCKS_REQUIRED(&crit_)` in line 111 of [webrtc absSendTime.h](https://chromium.googlesource.com/external/webrtc/+/master/webrtc/modules/remote_bitrate_estimator/remote_bitrate_estimator_abs_send_time.cc) for  `isBitrateImproving`

bgrozev: Ahh, got it. Ignore the rest of my comments on "synchronized"

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Alright, I believe it should be `incomingBitrate != 0`

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Alright, got it.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: see line 114 of [webrtc absSendTime.h](https://chromium.googlesource.com/external/webrtc/+/master/webrtc/modules/remote_bitrate_estimator/remote_bitrate_estimator_abs_send_time.h)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Okay, I didn't notice that. Thinking of this, I checked and saw this is supposed to be equivalent to 
`getLatestEstimate` in `RemoteBitrateEstimator`. In jitsi implementation, it returns `long` and [webrtc - line 63](https://chromium.googlesource.com/external/webrtc/+/master/webrtc/modules/remote_bitrate_estimator/include/remote_bitrate_estimator.h) returns `bool`. Although it is left to us to decide how to implement and use it. Thinking of the usage, I am going to make it return long. if true or valid, it `returns bitrateBps` else if false, it `returns -1`. It should be left for the calling method to check the value and use accordingly. let me know if this will be fine.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: In the `[webrtc RemoteBitrateEstimator.h - line 57](https://chromium.googlesource.com/external/webrtc/+/master/webrtc/modules/remote_bitrate_estimator/include/remote_bitrate_estimator.h)`, it is commented as   `// Removes all data for |ssrc|.` Am not sure if am to just remove the `ssrc` value from the list (which is redundant) here or if this is supposed to be equivalent to `removeMediaStream` in the `transportCCEngine`.

bgrozev: I think this should just remove the state for the particular SSRC (which as you say is redundant). The removeMediaStream in TransportCCEngine is unrelated.

jcchuks: Alright

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Okay

bgrozev:
jenkins test this again

gpolitis:
It may very well be the case that I don't remember correctly, but didn't we agree to introduce the ABS sendtime remote bitrate estimator separately from TCC support? Here I see changes related to TCC.

jcchuks:
Changes in this commits depends on https://github.com/jitsi/libjitsi/pull/336

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.impl.neomedia.rtp.remotebitrateestimator;
+
+
+import org.ice4j.util.RateStatistics;
+import org.jitsi.impl.neomedia.rtp.TransportCCEngine;
```
bgrozev: Unused import

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+import org.jitsi.impl.neomedia.transform.*;
+import org.jitsi.service.neomedia.*;
+import org.jitsi.service.neomedia.rtp.*;
+import org.jitsi.util.*;
+import org.jitsi.util.concurrent.RecurringRunnable;
+import org.jitsi.util.function.Predicate;
```
bgrozev: Please use *

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+     * <p>
+     * XXX At some point ideally we would get rid of this ctor and all the
+     * inheritors will use the parametrized ctor. Also, we might want to move
+     * this check inside the <tt>TransformEngineChain</tt> so that we only make
+     * the check once per packet: The RTCP transformer is only supposed to
+     * (reverse) transform RTCP packets and the RTP transformer is only
+     * supposed to modify RTP packets.
+     */
+    public RemoteBitrateEstimatorAbsSendTime() {
+        super();
```
bgrozev: Why do we need this constructor?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+    }
+
+    /**
+     * Ctor.
+     *
+     * @param packetPredicate the <tt>PacketPredicate</tt> to use to match
+     *                        packets to (reverse) transform.
+     */
+    public RemoteBitrateEstimatorAbsSendTime(Predicate<ByteArrayBuffer> packetPredicate) {
+        super(packetPredicate);
```
bgrozev: Why do we need this constructor? The RBE transformer only works with RTP packets, so it should always call super(RTPPacketPredicate.INSTANCE)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+
+    /**
+     * {@inheritDoc}
+     * <p>
+     * The (default) implementation of {@code SinglePacketTransformer} does
+     * nothing.
+     */
+    @Override
+    public void close() {
+        super.close();
```
bgrozev: Why do we override if we're only calling super.close()?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+     * {@inheritDoc}
+     * <p>
+     * Reverse-transforms an array of packets by calling
+     * {@link #reverseTransform(RawPacket)} on each one.
+     *
+     * @param pkts
+     */
+    @Override
+    public RawPacket[] reverseTransform(RawPacket[] pkts) {
+        return super.reverseTransform(pkts);
```
bgrozev: Why do we override if we're only calling super?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+
+    /**
+     * Transforms a specific packet.
+     *
+     * @param pkt the packet to be transformed.
+     * @return the transformed packet.
+     */
+    @Override
+    public RawPacket transform(RawPacket pkt) {
+        return null;
```
bgrozev: This effectively drops any outgoing packets.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+     * {@inheritDoc}
+     * <p>
+     * Transforms an array of packets by calling {@link #transform(RawPacket)}
+     * on each one.
+     *
+     * @param pkts
+     */
+    @Override
+    public RawPacket[] transform(RawPacket[] pkts) {
+        return super.transform(pkts);
```
bgrozev: Why do we override if we're only calling super?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+     * Returns the number of milliseconds until this instance wants a worker
+     * thread to call {@link #run()}. The method is called on the same
+     * worker thread as Process will be called on.
+     *
+     * @return the number of milliseconds until this instance wants a worker
+     * thread to call {@link #run()}
+     */
+    @Override
+    public long getTimeUntilNextRun() {
+        return 0;
```
bgrozev: Why do we implement RecurringRunnable? Returning 0 here will cause Process() to be executed immediately (if the instance is registered with an executable).

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+    }
+
+    private enum ProbeResult
```
bgrozev: Please move class/enum definitions to the end of the file (after method definitions)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+    private Collection<Integer> ssrcs = new ArrayList<Integer>();
+    private final static int kTimestampGroupLengthMs = 5;
+    private final static int kAbsSendTimeFraction = 18;
+    private final static int kAbsSendTimeInterArrivalUpshift = 8;
+    private final static int kInterArrivalShift
+            =  kAbsSendTimeFraction + kAbsSendTimeInterArrivalUpshift;
+    private final static int kInitialProbingIntervalMs = 2000;
+    private final static int kMinClusterSize = 4;
+    private final static int kMaxProbePackets = 15;
+    private final static int kExpectedNumberOfProbes = 3;
```
bgrozev: Please move all static fields to the top of the file. The order is: statics, instance fields, constructors, methods, classes.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+        }
+        return keys;
+    }
+
+    private long ConvertMsTo24Bits(long timeMs)
+    {
+        long time24Bits = (long)(((timeMs << kAbsSendTimeFraction) + 500) /
+                1000) & 0x00FFFFFF;
+        return time24Bits;
+    }
```
bgrozev: Method definitions should go below the class fields and constructors (see above)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+    private AimdRateControl remoteRate  = new AimdRateControl();
+    private InterArrival interArrival;
+    private OveruseEstimator estimator;
+    private OveruseDetector detector;
+    private OverUseDetectorOptions options;
+    private RateStatistics incomingBitrate ;
+    boolean incomingBitrateInitialized;
+    ArrayList<Integer> recentPropagationDeltaMs;
+    ArrayList<Long> recentUpdateTimeMs;
+    private AbsSendTimeEngine absoluteSendTimeEngine = new AbsSendTimeEngine();
```
bgrozev: This explains why you needed the extension_id_for_debugging in the AbsSendTimeEngine. Don't create a new instance, get the instance from the MediaStream (you will need to expose MediaStreamImpl#absSendTimeEngine).

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+    {
+        cluster.sendMeanMs /= (double)cluster.count;
+        cluster.recvMeanMs /= (double)cluster.count;
+        cluster.meanSize /= cluster.count;
+        clusters.add(cluster);
+    }
+
+    public RemoteBitrateEstimatorAbsSendTime(RemoteBitrateObserver observer)
+    {
+        this.observer_ = observer;
```
bgrozev: Call super(RTPPacketPredicate.INSTANCE), so that reverseTransform is only passed RTP packets. Move the constructor to the proper place.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+                        + " ms, mean recv delta: " + cluster.recvMeanMs
+                        + " ms, num probes: " + cluster.count);
+                break;
+            }
+        }
+        return bestIt;
+    }
+
+    @Override
+    public void run(){}
```
bgrozev: Why do we need this?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+
+import java.util.*;
+
+/**
+ * webrtc.org abs_send_time implementation as of June 26, 2017.
+ * commit ID: 23fbd2aa2c81d065b84d17b09b747e75672e1159
+ * @author Julian Chukwu
+ */
+public class RemoteBitrateEstimatorAbsSendTime
+    extends SinglePacketTransformer
```
bgrozev: You could extend SinglePacketTransformerAdapter instead, to avoid having to implement a dummy transform()

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+     */
+    @Override
+    public RawPacket reverseTransform(
+            RawPacket packet)
+    {
+        if (absoluteSendTimeEngine
+                .hasAbsoluteSendTimeExtension(packet) == null) {
+            logger.warn( "RemoteBitrateEstimatorAbsSendTimeImpl: Incoming" +
+                    " packet is missing absolute send time extension!");
+        return null;
```
bgrozev: This drops the packet. This shouldn't happen just because it doesn't have the extension. I'd expect the wrapper to handle this logic.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+        long arrivalTimeMs,
+        long sendTime24bits,
+        long payloadSize,
+        long ssrc) {
+        /**
+         * @ToDo 1L << 24 should be 1L << 23 since we want to check if
+         * sendTime24bits is less than 24 bits
+         */
+        if (sendTime24bits < (1L << 24)){
+            logger.warn("Send Time not valid");
```
bgrozev: This check doesn't seem right. Valid values should be [0, 2^24-1], so the check should be something like if (sendTime24bits < 0 || sendTime24bits >= (1 << 24))

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+        }
+    }
+
+    void timeoutStreams(long nowMs)
+    {
+        synchronized (critSect) {
+            Iterator<Map.Entry<Long, Long>> itr = ssrcs_.entrySet().iterator();
+            while (itr.hasNext()) {
+                /**
+                 * @Todo ask if this is the best way to modify a treemap object
```
bgrozev: Yeah, this looks like the right way to modify a map. You can avoid the cast below if you declare "entry" to be of type Map.Entry<Long, Long>

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+
+    void removeStream(long ssrc)
+    {
+        synchronized (critSect)
+        {
+            ssrcs_.remove(ssrc);
+        }
+    }
+
+    long getlatestEstimate(List<Long> ssrcs,
```
bgrozev: Is the capitalization this way on purpose? Is this used anywhere?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,719 @@
...
+    }
+
+    @Override
+    public Collection<Integer> getSsrcs() {
+
+        synchronized (critSect)
+        {
+            for(Long ssrcValue : ssrcs_.keySet()){
+                Number value = ssrcValue;
+                ssrcs.add(value.intValue());
```
bgrozev: This keeps adding to the same ArrayList (this.ssrcs). You probably want a new instance.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Okay, method can be removed.

jitsi-developers:
Martin

On 17 July 2017 at 22:54, bgrozev <notifications@github.com> wrote:

> *@bgrozev* commented on this pull request.
> ------------------------------
>
> In src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/
> RemoteBitrateEstimatorAbsSendTime.java
> <https://github.com/jitsi/libjitsi/pull/334#discussion_r127822576>:
>
> > +                        .getAbsSendTime(packet),
> +        if (sendTime24bits < (1L << 24)){
> +            logger.warn("Send Time not valid");
>
> This check doesn't seem right. Valid values should be [0, 2^24-1], so the
> check should be something like if (sendTime24bits < 0 || sendTime24bits >=
> (1 << 24))
>
>
> Personally, I would tend to write as:
        if (0 <= sendTime24bits && sendTime24bits < (1<<24))
        {
            // OK
        }
        else
        {
            logger.warn("Send Time not valid");
        }

I find combining multiple logical clauses hard to read, and the above is
the simplest for me to understand.  Others may disagree, and find it
needlessly verbose (but I'd like the author to actively decide "nope, not
going to do that").

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Fixed

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Fixed. Note Taken

jcchuks:
Hi Martin, Thanks for the suggestion, yeah that is another way to write it though I still want to be able to go ahead and do the computations no matter the condition. I find the OR variation easy to read too. I could also consider a variation of what you wrote to eliminate the else block and still perform the computation.
```
if ( ! (0 <= sendTime24bits && sendTime24bits < (1<<24) ) ){
         logger.warn("Send Time not valid");
} 
```
which reads -  if sendTime24Bits is not within the specified range, log "Send Time not Valid". But, I am fine with the ORs.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+            (1 << kInterArrivalShift) ;
+
+    private final Object critSect = new Object();
+    private Collection<Integer> ssrcs = new ArrayList<Integer>();
+    private  TreeMap<Long,Long> ssrcs_ = new TreeMap<Long, Long>();
+    private  ArrayList<Probe> probes_ = new ArrayList<>();
+    private  long totalProbesReceived;
+    private  long firstPacketTimeMs;
+    private  long lastUpdateMs;
+    public RemoteBitrateObserver observer_;
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+    private AimdRateControl remoteRate  = new AimdRateControl();
+    private InterArrival interArrival;
+    private OveruseEstimator estimator;
+    private OveruseDetector detector;
+    private RateStatistics incomingBitrate ;
+    boolean incomingBitrateInitialized;
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+
+    private long ConvertMsTo24Bits(long timeMs)
+    {
+        long time24Bits = (long)(((timeMs << kAbsSendTimeFraction) + 500) /
+                1000) & 0x00FFFFFF;
+        return time24Bits;
+    }
+
+
+    boolean IsWithinClusterBounds(long sendDeltaMs, Cluster clusterAggregate)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+        if(clusterAggregate.count == 0)
+        {
+            return true;
+        }
+        double clusterMean = clusterAggregate.sendMeanMs /
+                (double)clusterAggregate.count;
+        return  Math.abs((double)sendDeltaMs - clusterMean) < 2.5f;
+    }
+
+    void addCluster(List<Cluster> clusters, Cluster cluster)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+    {
+        cluster.sendMeanMs /= (double)cluster.count;
+        cluster.recvMeanMs /= (double)cluster.count;
+        cluster.meanSize /= cluster.count;
+        clusters.add(cluster);
+    }
+
+    void computeClusters(List<Cluster> clusters)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+        }
+        if(current.count >= kMinClusterSize)
+            addCluster(clusters, current);
+    }
+
+    /**
+     * @param clusters
+     * @returns a cluster that shows the best probe
+     */
+    Cluster findBestProbe(List<Cluster> clusters)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+                        + " bps. Mean send delta: " + cluster.sendMeanMs
+                        + " ms, mean recv delta: " + cluster.recvMeanMs
+                        + " ms, num probes: " + cluster.count);
+                break;
+            }
+        }
+        return bestIt;
+    }
+
+    ProbeResult processClusters(long nowMs)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+            }
+            // Not probing and received non-probe packet,
+            // or finished with current set  of probes.
+            if (clusters.size() >= kExpectedNumberOfProbes)
+                probes_.clear();
+            return ProbeResult.kNoUpdate;
+        }
+    }
+
+    boolean isBitrateImproving(int newBitrateBps)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+    {
+        logger.info("Using RemoteBitrateEstimatorAbsSendTime: Instantiating.");
+
+        incomingPacketInfo(System.currentTimeMillis(), absoluteSendTimeEngine
+                .getAbsSendTime(packet), packet.getPayloadLength(),
+                packet.getSSRC());
+        return packet;
+    }
+
+    void incomingPacketInfo(
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+            }
+        }
+        if (updateEstimate)
+        {
+            lastUpdateMs = nowMs;
+            observer_.onReceiveBitrateChanged(getSsrcs(), targetBitrateBps);
+        }
+    }
+
+    void timeoutStreams(long nowMs)
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+
+    void OnRttUpdate(long avg_rtt_ms,
+                     long max_rtt_ms)
+    {
+        synchronized (critSect) {
+            remoteRate.setRtt(avg_rtt_ms);
+        }
+    }
+
+    void removeStream(long ssrc)
```
bgrozev: Why do we have two removeStream methods?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+     */
+    @Override
+    public RawPacket reverseTransform(
+            RawPacket packet)
+    {
+        logger.info("Using RemoteBitrateEstimatorAbsSendTime: Instantiating.");
+
+        incomingPacketInfo(System.currentTimeMillis(), absoluteSendTimeEngine
+                .getAbsSendTime(packet), packet.getPayloadLength(),
+                packet.getSSRC());
```
bgrozev: You need getSSRCAsLong() here

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+
+    @Override
+    public Collection<Integer> getSsrcs() {
+
+        synchronized (critSect)
+        {
+            ssrcs = new ArrayList<>();
+            for(Long ssrcValue : ssrcs_.keySet()){
+                Number value = ssrcValue;
+                ssrcs.add(value.intValue());
```
bgrozev: Why do we save this instance (ssrcs) in a class field? It isn't used anywhere else.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+     * Removes all data for <tt>ssrc</tt>.
+     *
+     * @param ssrc
+     */
+    @Override
+    public void removeStream(int ssrc)
+    {
+        synchronized (critSect) {
+            try {
+                ssrcs_.remove(ssrc);
```
bgrozev: You need to handle the conversion from int to long here: ssrcs_.remove(ssrc & 0xFFFF_FFFFL)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+            return (int) (this.meanSize * 8 * 1000 / sendMeanMs);
+        }
+
+        public int getRecvBitrateBps() {
+           // RTC_CHECK_GT(this.recvMeanMs, 0.0f);
+            return (int) (this.meanSize * 8 * 1000 / this.recvMeanMs);
+        }
+    }
+
+    public class Probe
```
bgrozev: private

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,601 @@
...
+        // Called from both the configuration thread and the network thread.
+        // Shouldn't be called from the network thread in the future.
+        synchronized (critSect)
+        {
+            remoteRate.setMinBitrate(minBitrateBps);
+        }
+
+    }
+
+    public class Cluster {
```
bgrozev: private

bgrozev:
> It may very well be the case that I don't remember correctly, but didn't we agree to introduce the ABS sendtime remote bitrate estimator separately from TCC support? Here I see changes related to TCC.

If that's still the case can you please point to the TCC-related stuff?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: SSRC is not used directly. I am using the method - getSsrc as a wrapper to convert and return SSRC values from Collection<Long> to type Collection<Integer>.

bgrozev: Yeah, that's fine. But why do you save the result in the "ssrcs" field of RemoteBitrateEstimatorAbsSendTime?

jcchuks: The only function of ssrc is to serve as a container for getting ssrc values anytime the getSsrc method is called, I declared it as a class member instead of inside the getSsrc method  to avoid re-declaring that variable anytime getSsrc() is called.

I have moved the declaration within the method. Though I was thinking there might be an added advantage or improvement by declaring once (as a class property) and assigning multiple times over declaring multiple times and assigning multiple times. Please correct me if am wrong. Thanks.

bgrozev: I am not aware of any advantage of this approach. Perhaps if you re-used the same List instance there would be a little less garbage collection going on, but I wouldn't worry about it (and I would advice confirming the performance gain before using a more complex solution).

jcchuks:
Alright 


Sent from Yahoo Mail for iPhone


On Saturday, July 22, 2017, 7:29 PM, bgrozev <notifications@github.com> wrote:


@bgrozev commented on this pull request.

In src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java:
> +     * Returns the estimated payload bitrate in bits per second if a valid
+     * estimate exists; otherwise, <tt>-1</tt>.
+     *
+     * @return the estimated payload bitrate in bits per seconds if a valid
+     * estimate exists; otherwise, <tt>-1</tt>
+     */
+
+    @Override
+    public Collection<Integer> getSsrcs() {
+
+        synchronized (critSect)
+        {
+            ssrcs = new ArrayList<>();
+            for(Long ssrcValue : ssrcs_.keySet()){
+                Number value = ssrcValue;
+                ssrcs.add(value.intValue());

I am not aware of any advantage of this approach. Perhaps if you re-used the same List instance there would be a little less garbage collection going on, but I wouldn't worry about it (and I would advice confirming the performance gain before using a more complex solution).

bgrozev:
Jenkins, test this again please

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -0,0 +1,593 @@
...
+ * commit ID: 23fbd2aa2c81d065b84d17b09b747e75672e1159
+ * @author Julian Chukwu
+ */
+public class RemoteBitrateEstimatorAbsSendTime
+    extends SinglePacketTransformerAdapter
+    implements RemoteBitrateEstimator
+{
+    //@Todo Ask for alternative to resolve import conflict between
+    //org.jitsi.util.* and org.ice4j.util.* when importing Logger
+    // and RateStatistics. For now, see below.
```
bgrozev: Using the full name is fine. In these cases, as an exception, we import the specific class that we need (import org.jitsi.util.Logger).

PR #336, opened 2017-07-14
Add check for AbsSendTimeExtension and method to retrieve AbsSendTime

jcchuks:


src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
```diff
@@ -129,8 +130,66 @@ private void setTimestamp(byte[] buf, int off)
...
     {
         extensionID = id;
     }
+
+    /**
+     *
+     * @param packet is the RawPacket
+     * @returns the Absolute Send Time header extension if present
+     */
+    public HeaderExtension hasAbsoluteSendTimeExtension(RawPacket packet)
```
bgrozev: Can we rename this to getAbsoluteSendTimeExtension? Do you need to use this from outside the class? If not, it should be private.

src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
```diff
@@ -129,8 +130,66 @@ private void setTimestamp(byte[] buf, int off)
...
+    {
+
+        if(packet.getExtensionBit() )
```
bgrozev: We should also check if we have a valid ID here (RawPacket.getHeaderExtension() will return null if we ask for -1, but it will still loop through the extensions which isn't necessary).

src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
```diff
@@ -129,8 +130,66 @@ private void setTimestamp(byte[] buf, int off)
...
+        return null;
+    }
+
+    /**
+     * @ToDo findout the prefered way of setting {@Link extensionID}
+     * returns AbsSendTime Extension ID
+     * @return
+     */
+    public byte getAbsSendTimeExtensionID(){
+        return extensionID == -1 ? extensionID_for_debugging : extensionID ;
```
bgrozev: Can we please remove extensionID_for_debugging? When testing with jitsi-meet abs-send-time is configured with id=3, so you can already test it.

src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
```diff
@@ -129,8 +130,66 @@ private void setTimestamp(byte[] buf, int off)
...
+        long absSendTime = -1L;
+        HeaderExtension header  = hasAbsoluteSendTimeExtension(pkt);
+        if (header != null)
+        {
+            //offSet is the byte index to read from
+            int offSet = header.getOffset() + 1;
+            if (header.getExtLength() == EXT_LENGTH) {
+                absSendTime = (header.getBuffer()[offSet] & 0xFF) << 16
+                        | (header.getBuffer()[offSet + 1] & 0xFF) << 8
+                        | (header.getBuffer()[offSet + 2] & 0xFF);
```
bgrozev: You can use the RTPUtils.readUint24AsInt() utility

src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
```diff
@@ -129,8 +130,52 @@ private void setTimestamp(byte[] buf, int off)
...
     }
+
+    /**
+     *
+     * @param packet is the RawPacket
+     * @returns the Absolute Send Time header extension if present
+     */
+    public HeaderExtension getAbsoluteSendTimeExtension(RawPacket packet)
+    {
+        if( packet.getExtensionBit() && packet.getHeaderExtensionType()
```
bgrozev: We should check whether the AbsSendTimeEngine is configured with an extension ID (e.g. getExtensionID() != -1).

bgrozev: The header extension type is not related to the extension ID.

PR #337, opened 2017-07-14
Logic to switch between RemoteBitrateEstimators

jcchuks:
RemoteBitrateSelector - Logic to switch between RemoteBitrateEstimatorrAbsSendTime and RemoteBitrateEstimatorSingleStream

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
+++ b/src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
@@ -15,6 +15,7 @@
  */
 package org.jitsi.impl.neomedia.rtp.remotebitrateestimator;
 
+import org.jitsi.impl.neomedia.*;
 import org.jitsi.impl.neomedia.transform.*;
 import org.jitsi.service.configuration.*;
 import org.jitsi.service.libjitsi.*;
@@ -46,6 +47,7 @@ public class RemoteBitrateEstimatorSelector
     private AbsSendTimeEngine absSendTimeEngine;
     private long packetsSinceAbsoluteSendTime;
     private int minBitrateBps;
+    private VideoMediaStreamImpl stream;
     private SinglePacketTransformer packetTransformer;
     private final Object critSect = new Object();
     private RemoteBitrateObserver observer;
@@ -56,20 +58,16 @@ public class RemoteBitrateEstimatorSelector
      */
     private boolean enableUseOfAbsSendTime;
 
-    public RemoteBitrateEstimatorSelector(RemoteBitrateObserver observe, AbsSendTimeEngine _absSendTimeEngine)
+    public RemoteBitrateEstimatorSelector(RemoteBitrateObserver observe, VideoMediaStreamImpl stream)
     {
         this.observer = observe;
-
+        this.stream = stream;
         /**
          * RemoteBitrateEstimatorSingleStream is the Default BitrateEstimator
          */
         this.packetTransformer
                 = new RemoteBitrateEstimatorSingleStream(observe);
         this.minBitrateBps = 0;
-
-        //@Todo ask how to retrieve the absSendTimeEngine from MediaStream
-        //Temporary fix
-        this.absSendTimeEngine = _absSendTimeEngine == null ? new AbsSendTimeEngine(): _absSendTimeEngine;
         this.packetsSinceAbsoluteSendTime = 0;
         setAbsSendTimeUsageConfiguration();
     }
@@ -188,6 +186,10 @@ public class RemoteBitrateEstimatorSelector
          * @Todo Find out why having synchronized here makes sense since it
          * exists in PickEstimatorFromHeader.
          */
+        if (this.absSendTimeEngine == null){
+            this.absSendTimeEngine = stream.getAbsSendTimeEngine();
+        }
+
         synchronized (critSect)
         {
             if(enableUseOfAbsSendTime)
--- a/src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
+++ b/src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
@@ -4,6 +4,7 @@ import org.jitsi.impl.neomedia.transform.*;
 import org.jitsi.service.neomedia.*;
 import org.jitsi.service.neomedia.rtp.*;
 import org.jitsi.util.*;
+import org.jitsi.util.concurrent.*;
 
 import java.util.*;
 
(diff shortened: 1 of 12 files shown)


src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
```diff
@@ -0,0 +1,268 @@
...
+     *
+     * @return the number of milliseconds until this instance wants a worker
+     * thread to call {@link #run()}
+     */
+    @Override
+    public long getTimeUntilNextRun() {
+        long timeUntilNextRun  =
+                (packetTransformer instanceof RecurringRunnable) ?
+             ((RecurringRunnable) packetTransformer)
+                     .getTimeUntilNextRun() : -1;
```
bgrozev: Returning -1 here will result in the executor running this immediately (there are no special semantics for -1, it just executes as long as getTimeUntilNextRun <= 0). I think the easiest solution would be to run it at the same interval as RBESingleStream (500ms). Another option would be to only register with the executor while we're using RBESingleStream, and then de-register when we switch to RBEAbsSendTime.

jcchuks: Updated it to return 500Ms. I think it is a better choice. If RBEAbsSendTime should for some reason implement the getTimeUntilNextRun method in the future. Nothing will need to be changed in the Selector.

gpolitis:
This file needs to be reformatted to meet the [Jitsi Coding Convention](https://desktop.jitsi.org/Documentation/CodeConvention). IntelliJ IDEA lets you [reformat source code](https://www.jetbrains.com/help/idea/reformatting-source-code.html) to meet the requirements of your code style, but it looks like you haven't configured it to do so. I have a Jitsi.xml [here](https://gist.github.com/gpolitis/19b75e4814638f8153ab2369a8599549) that you can import in IntelliJIDEA to configure the proper coding style.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
```diff
@@ -0,0 +1,269 @@
...
+        {
+            ((Runnable)packetTransformer).run();
+        }
+    }
+
+    @Override
+    public void onRttUpdate(long avgRttMs, long maxRttMs) {
+        if(packetTransformer instanceof CallStatsObserver){
+            ((CallStatsObserver)packetTransformer)
+                    .onRttUpdate(avgRttMs,maxRttMs);
```
gpolitis: Note that the `RemoteBitrateEstimatorAbsSendTime` does not currently implement the `CallStatsObserver`.

gpolitis:
Can we simplify the `RemoteBitrateEstimatorSingleStream` and make it _not_ implement the `TransformEngine` interface?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
```diff
@@ -0,0 +1,269 @@
...
+     * The general contract of the method <code>run</code> is that it may
+     * take any action whatsoever.
+     *
+     * @see Thread#run()
+     */
+    @Override
+    public void run() {
+        if (packetTransformer instanceof Runnable)
+        {
+            ((Runnable)packetTransformer).run();
```
gpolitis: Note that the `RemoteBitrateEstimatorAbsSendTime` does not currently implement the `Runnable`. We should probably _not_ have both the selector and the bitrate estimator implement the `RecurringRunnable` interface.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
```diff
@@ -0,0 +1,269 @@
...
+                TransformEngine
+{
+
+    public final static String ENABLE_USE_OF_ABS_SEND_TIME =
+            "org.jitsi.impl.neomedia.rtp.ENABLE_ABS_SEND_TIME";;
+    public static final long kTimeOffsetSwitchThreshold = 30;
+    private static final Logger logger = Logger
+            .getLogger(RemoteBitrateEstimatorSelector.class);
+    private static long defaultTimeUntilNextRunMs = 500;
+    private boolean usingAbsoluteSendTime;
```
gpolitis: We usually declare configuration parameters as final.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /**
+     * Selects either to use RBE_abs_send_time or RBE_single_stream
+     */
+    private void PickEstimator()
+    {
+        synchronized (critSect)
+        {
+            if (usingAbsoluteSendTime)
```
gpolitis: The `synchronized` can be inside the `if` statement, so we don't block if we're not using the ABSRBE.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
```diff
@@ -0,0 +1,269 @@
...
+        }
+        //@Todo set default value to false. Allow true for test.
+        enableUseOfAbsSendTime
+                = cfg.getBoolean(ENABLE_USE_OF_ABS_SEND_TIME, false);
+    }
+
+    /**
+     * Selects either to use RBE_abs_send_time or RBE_single_stream
+     */
+    private void PickEstimator()
```
gpolitis: Capitalization.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
```diff
@@ -0,0 +1,269 @@
...
+        /**
+         * RemoteBitrateEstimatorSingleStream is the Default BitrateEstimator
+         */
+        this.packetTransformer
+                = new RemoteBitrateEstimatorSingleStream(observe);
+        this.minBitrateBps = 0;
+        this.packetsSinceAbsoluteSendTime = 0;
+        setAbsSendTimeUsageConfiguration();
+    }
+    private void PickEstimatorFromHeader(RawPacket packet) {
```
gpolitis: Capitalization.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
jcchuks: Does this mean the synchronized construct can be omitted from the `else ` block?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSelector.java
jcchuks: Currently, only `RemoteBitrateEstimatorSelector ` and `RemoteBitrateEstimatorSingleStream ` implements the `RecurringRunnable ` interface. By your comments, do you mean RecurringRunnable should be removed.

PR #338, opened 2017-07-14
Hooks up RemoteBitrateSelector to the TransportChain.

jcchuks:


gpolitis:
This needs to be part of the PR that introduces the RBE selector.

PR #340, opened 2017-07-19
Unbreaks H264 support in the bridge.

gpolitis:
This is providing https://github.com/jitsi/jitsi-videobridge/pull/504

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -3668,7 +3668,7 @@ public class MediaStreamImpl
             // wrong results if the payload is encrypted.
         }
     
-        REDBlock redBlock = getPayloadBlock(buf, off, len);
+        REDBlock redBlock = getPrimaryREDBlock(buf, off, len);
         if (redBlock == null || redBlock.getLength() == 0)
         {
             return -1;
@@ -3733,7 +3733,7 @@ public class MediaStreamImpl
             // wrong results if the payload is encrypted.
         }
 
-        REDBlock redBlock = getPayloadBlock(buf, off, len);
+        REDBlock redBlock = getPrimaryREDBlock(buf, off, len);
         if (redBlock == null || redBlock.getLength() == 0)
         {
             return -1;
@@ -3772,7 +3772,7 @@ public class MediaStreamImpl
     {
         if (frameMarkingsExtensionId == -1)
         {
-            REDBlock redBlock = getPayloadBlock(pkt);
+            REDBlock redBlock = getPrimaryREDBlock(pkt);
             if (redBlock != null && redBlock.getLength() != 0)
             {
                 final byte vp9PT = getDynamicRTPPayloadType(Constants.VP9),
@@ -3824,7 +3824,7 @@ public class MediaStreamImpl
             // wrong results if the payload is encrypted.
         }
 
-        REDBlock redBlock = getPayloadBlock(pkt);
+        REDBlock redBlock = getPrimaryREDBlock(pkt);
         if (redBlock == null || redBlock.getLength() == 0)
         {
             return false;
@@ -3884,7 +3884,7 @@ public class MediaStreamImpl
             // wrong results if the payload is encrypted.
         }
 
-        REDBlock redBlock = getPayloadBlock(pkt);
+        REDBlock redBlock = getPrimaryREDBlock(pkt);
         if (redBlock == null || redBlock.getLength() == 0)
         {
             return false;
@@ -3942,7 +3942,7 @@ public class MediaStreamImpl
             // wrong results if the payload is encrypted.
         }
 
-        REDBlock redBlock = getPayloadBlock(buf, off, len);
+        REDBlock redBlock = getPrimaryREDBlock(buf, off, len);
         if (redBlock == null || redBlock.getLength() == 0)
         {
             return false;
@@ -4007,9 +4007,9 @@ public class MediaStreamImpl
      * {@inheritDoc}
      */
     @Override
-    public REDBlock getPayloadBlock(ByteArrayBuffer baf)
+    public REDBlock getPrimaryREDBlock(ByteArrayBuffer baf)
     {
-        return getPayloadBlock(
+        return getPrimaryREDBlock(
             baf.getBuffer(), baf.getOffset(), baf.getLength());
     }
 
@@ -4023,7 +4023,7 @@ public class MediaStreamImpl
      * @return the {@link REDBlock} that contains the payload of the packet
      * passed in as a parameter, or null if the buffer is invalid.
      */
-    public REDBlock getPayloadBlock(byte[] buf, int off, int len)
+    public REDBlock getPrimaryREDBlock(byte[] buf, int off, int len)
     {
         if (buf == null || buf.length < off + len
             || len < RawPacket.FIXED_HEADER_SIZE)
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -3697,7 +3697,8 @@ public class MediaStreamImpl
         }
         else
         {
-            logger.warn("Method not implemented");
+            // XXX not implementing temporal layer detection should not break
+            // things.
             return -1;
         }
     }
@@ -3751,8 +3752,43 @@ public class MediaStreamImpl
         }
         else
         {
-            logger.warn("Method not implemented");
+            // XXX not implementing temporal layer detection should not break
+            // things.
             return -1;
+        }
+    }
+
+    /**
+     * Returns a boolean that indicates whether or not our we're able to detect
+     * the frame boundaries for the codec of the packet that is specified as an
+     * argument.
+     *
+     * @param pkt the {@link RawPacket} that holds the RTP packet.
+     *
+     * @return true if we're able to detect the frame boundaries for the codec
+     * of the packet that is specified as an argument, false otherwise.
+     */
+    public boolean supportsFrameBoundaries(RawPacket pkt)
+    {
+        if (frameMarkingsExtensionId == -1)
+        {
+            REDBlock redBlock = getPayloadBlock(pkt);
+            if (redBlock != null && redBlock.getLength() != 0)
+            {
+                final byte vp9PT = getDynamicRTPPayloadType(Constants.VP9),
+                    vp8PT = getDynamicRTPPayloadType(Constants.VP8),
+                    pt = redBlock.getPayloadType();
+
+                return vp9PT == pt || vp8PT == pt;
+            }
+            else
+            {
+                return false;
+            }
+        }
+        else
+        {
+            return pkt.getHeaderExtension(frameMarkingsExtensionId) != null;
         }
     }
 
@@ -3775,10 +3811,6 @@ public class MediaStreamImpl
             return false;
         }
 
-        byte[] buf = pkt.getBuffer();
-        int off = pkt.getOffset();
-        int len = pkt.getLength();
-        
         if (frameMarkingsExtensionId != -1)
         {
             RawPacket.HeaderExtension fmhe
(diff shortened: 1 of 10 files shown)


src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -3751,12 +3752,47 @@ public int getSpatialID(byte[] buf, int off, int len)
...
+        if (frameMarkingsExtensionId == -1)
+        {
+            REDBlock redBlock = getPayloadBlock(baf);
+            if (redBlock != null && redBlock.getLength() != 0)
+            {
+                final byte vp9PT = getDynamicRTPPayloadType(Constants.VP9),
+                    vp8PT = getDynamicRTPPayloadType(Constants.VP8),
+                    pt = redBlock.getPayloadType();
+
+                return vp9PT == pt || vp8PT == pt;
```
bbaldino: should there be some h264 support here?  or you're thinking that will have to fall under the frameMarkingExtension scenario?  could't timestamp be used for h264?  or maybe this is only given a single packet.

bbaldino: also wondering if there's a different place where we could do the codec comparison that might be more generic (a utility function somewhere or something)?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -3977,6 +4004,16 @@ public TransformEngineChain getTransformEngineChain()
     }
 
     /**
+     * {@inheritDoc}
+     */
+    @Override
+    public REDBlock getPayloadBlock(ByteArrayBuffer baf)
```
bbaldino: should this be called `getREDBlock` ? (same for the method it calls inside as well)?  'payload' seems generic but this seems to return RED specifically

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
```diff
@@ -507,8 +507,30 @@ public static int getTemporalLayerIndex(byte[] buf, int off, int len)
...
          * <tt>offset</tt> in <tt>input</tt>. The size is between 1 and 6.
          *
+         * @param baf the <tt>ByteArrayBuffer</tt> that holds the VP8 payload
+         * descriptor.
+         *
+         * @return The size in bytes of the Payload Descriptor at offset
+         * <tt>offset</tt> in <tt>input</tt>, or -1 if the input is not a valid
+         * VP8 Payload Descriptor. The size is between 1 and 6.
+         */
+        public static int getSize(ByteArrayBuffer baf)
```
bbaldino: this looks like something that would fit into a more generic utils class, this isn't specific to vp8 depacketization, is it?

src/org/jitsi/impl/neomedia/rtp/FrameDesc.java
```diff
@@ -93,6 +103,18 @@
     }
 
     /**
+     * Gets a boolean that indicates whether or not we can read the frame
+     * boundaries of this frame.
+     *
+     * @return true if we're able to read the frame boundaries of this frame,
+     * false otherwise.
+     */
+    boolean supportsFrameBoundaries()
```
bbaldino: this has lower case `boolean` but the declaration has upper case `Boolean`

bgrozev: Yeah, is that gonna throw an NPE if called before #supportsFrameBounderies is re-initialized with a non-null?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
gpolitis: We don't have frame boundaries detection for H264; nor we need it, as it's only useful for SVC.

src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
gpolitis: This is merely adding an overloading an existing getSize() method, I'd rather not change the architecture + it's specific to VPX payload descriptor packetization.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -280,55 +280,84 @@ else if (tsDiff >= (-(1L << 30) & 0xFFFFFFFFL))
...
-                else if (snDiff < 3 || snDiff > (-4 & 0xFFFF))
+            }
+
+            if (guessed)
+            {
+                if (logger.isDebugEnabled())
                 {
-                    logger.warn("Frame corruption or packets that are out of" +
-                        " order detected.");
+                    logger.debug(
```
bgrozev: Is the above only refactoring (except for the check for withFrameBoundaries), or does it change the logic?

gpolitis: It doesn't change the idea, but it changes a few things. I was hoping you could have a closer look at some point.

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
gpolitis: sure, no problem. Some changes to the JVB where necessary: https://github.com/jitsi/jitsi-videobridge/pull/504/files

src/org/jitsi/impl/neomedia/rtp/FrameDesc.java
gpolitis: The field is now a final boolean.

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -3751,12 +3752,47 @@ public int getSpatialID(byte[] buf, int off, int len)
...
+                return vp9PT == pt || vp8PT == pt;
+            }
+            else
+            {
+                return false;
+            }
+        }
+        else
+        {
+            return true;
```
bgrozev: Should we also check whether the packet has a framemarkings extension? Here we only check whether it was signaled.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -280,55 +280,84 @@ else if (tsDiff >= (-(1L << 30) & 0xFFFFFFFFL))
...
+            boolean withFrameBoundaries =
+                a.supportsFrameBoundaries() && b.supportsFrameBoundaries();
+
+            if (withFrameBoundaries)
             {
-                if (snDiff == 2)
+                if (bFirstSeqNum != -1 || aLastSeqNum != -1)
                 {
-                    if (aLastSeqNum == -1)
+                    if (snDiff == 2 || snDiff == (-2 & 0xFFFF))
```
bgrozev: What is the idea behind -2 here? It seems to me that if the diff is -2 then "a" comes after "b", and we cannot apply the logic (setting a's end for example). What am I missing?

bgrozev: Brain dump: for VPX codecs with PictureID we could find the start/end even with diff>2 (if PictureIDDiff == 1)

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -280,55 +280,84 @@ else if (tsDiff >= (-(1L << 30) & 0xFFFFFFFFL))
...
             }
             else
             {
-                if (snDiff == 3)
+                if (bFirstSeqNum != -1 || aLastSeqNum != -1)
                 {
-                    bFirstSeqNum = (bMinSeqNum - 1) & 0xFFFF;
-                    aLastSeqNum = (aMaxSeqNum + 1) & 0xFFFF;
-                    if (logger.isDebugEnabled())
+                    if (snDiff == 1 || snDiff == (-1 & 0xFFFF))
```
bgrozev: Doesn't this apply more broadly, e.g. regardless of whether "bFirstSeqNum != -1 || aLastSeqNum != -1" is true or not? If diff==1, then we know both the end of a and the start of b.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -280,55 +280,84 @@ else if (tsDiff >= (-(1L << 30) & 0xFFFFFFFFL))
...
+
+                        guessed = true;
                     }
+                }
+                else // (bFirstSeqNum == -1 && aLastSeqNum == -1)
+                {
+                    if (snDiff == 2 || snDiff == (-2 & 0xFFFF))
+                    {
+                        a.setEnd(aMaxSeqNum & 0xFFFF);
+                        b.setStart(bMinSeqNum & 0xFFFF);
```
bgrozev: I don't understand this. If diff==2 and (bFirstSeqNum == -1 && aLastSeqNum == -1), then there is 1 packet between aLastSeen and bFirstSeen. And we don't know whether this packet belongs to a or to b, or is a separate frame of it's own (since in this if branch there is no support for frame boundaries, which means that e.g. aLastSeen could be the end of a even if aLastSeqNum == 1).

bbaldino:
is there any way we could get a helper method(s) for the masking?  it'd make things much easier to read/understand

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
gpolitis: Thanks @bgrozev, you're right, we cannot apply the logic if the diff is -2.

gpolitis: I like your idea about including the picture ID in our heuristics, but.. we don't have it in FrameDesc and I feel it doesn't belong there. I feel we may need to subclass it into VPXFrameDesc and H264FrameDesc and move the heuristics logic in there.

I added a note about your idea, but I'd rather not implement it now.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
gpolitis: You're right, if diff==1, then we know both the end of a and the start of b. It is also true that, if diff ==1, then we know that we must have seen the end (because the packet is marked), so `aLastSeqNum != -1`.

The reason why I prefer to keep the check is because of in-band probing. In that case a frame as identified by the RTP timestamp can be longer than the frame as identified by the marker bit and/or the end-of-frame indication inside the payload.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
gpolitis: You're right.

gpolitis:
@bbaldino sure, you mean something like this?

    boolean function seqNumDiffIsThree(int snDiff)
    {
      return snDiff == 3 || snDiff == (-3 & 0xFFFF)
    }

bbaldino:
@gpolitis oops, sorry...no i meant the masking itself, e.g.
`(-3 & 0xFFFF)`
To outsiders (including myself) it's confusing to read this and I'm not sure why it's done.  I'm wondering if we could use a named helper that would make it more clear what's being done & why?

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -280,55 +280,92 @@ else if (tsDiff >= (-(1L << 30) & 0xFFFFFFFFL))
...
                     {
-                        logger.debug(
-                            "Guessed frame start=" + bFirstSeqNum
-                                + ",end=" + aLastSeqNum);
-                    }
+                        if (aLastSeqNum == -1)
+                        {
+                            a.setEnd(aMaxSeqNum & 0xFFFF);
+                        }
+                        else
```
bgrozev: I don't think there's a problem here, but it seems strange: If (aLastSeqNum!=-1 &&  bFirstSeqNum!=-1), we update "b" and only "b".

bgrozev: Nope, I'm wrong. In that case the method has returned early on.

gpolitis:
@bbaldino this is now merged, but I'll make sure to add some constants and appropriate comments in a future PR.

PR #344, opened 2017-07-25
an attempt at refactoring RTPEncodingDesc::applyFrameBoundsHeuristics…

bbaldino:
… to make it more readable


bbaldino:
@bgrozev @gpolitis let me know what you think of this take.  i tried to map variable names/helper functions to what was logically being done so that `applyFrameBoundsHeuristics` reads a bit easier to the non-initiated (such as myself).  note that i do have a couple open questions in there as comments that i don't intend on committing, but wanted to get answers from you guys.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -249,124 +249,128 @@ public RTPEncodingDesc(
...
+
+        // We have some codecs (VPX) for which we can inspect the payload
+        // to deduce the start/end of olderFrame frame. This has an impact on the
+        // distances between the frames that allow us to guess frame
+        // boundaries.
+        //FIXME(brian): why does this allow us to infer anything extra from the sequence number
+        // differences?  we don't leverage frame boundary markings in any way when we deduce
+        // first/last sequence numbers.  does it just mean that these frames would've already
+        // (conclusively) had the start/end seq numbers set?  if so, we wouldn't need to
+        // call this method at all for those frames, right?
```
bbaldino: @bgrozev or @gpolitis i'm sure i'm missing something here that one of you guys can clear up.

bbaldino:
also...there was no intended change in logic here, just renaming variables and using existing/adding new helper functions.  looks like the way the diff shows up doesn't illustrate that very well, but that was the intention.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
bgrozev: I think you are right, if diff=2. But do could extract some extra info if diff>=3. For example if diff=3, we have the end of a, and we don't have the start of b. In this case the start of b is either bMin-1 or bMin-2, but we don't know which. If a and b have subsequent PictureIDs we can decude that b starts at bMin-2 (and if PictureID(b) = PictureID(a) + 2 we can deduce that b starts at bMin-1. And that extends for diff>3. 
I'm not sure it's worth implementing this, as it adds extra complexity for quite a rare case.

bbaldino: thanks boris.  but we don't leverage anything like picture id here, so i was trying to figure out why we have the if block for `if (framesSupportFrameBoundaries)` at all if we don't have the picture id to work with?

bbaldino: in fact, here if the diff is 3 we make an assumption but without using any sort of frame marking at all so it seems like a total guess...i don't get why we can make that guess if frame markings are supported but we don't know what they are

src/org/jitsi/util/RTPUtils.java
```diff
@@ -180,6 +181,28 @@ public static int readUint24AsInt(byte[] buffer, int offset)
     }
 
     /**
+     * Returns the given integer masked to 16 bits
+     * @param value the integer to mask
+     * @return the value, masked to only keep the lower
+     * 16 bits
+     */
+    public static int as16Bits(int value)
```
bbaldino: wish generics could be used here but they don't work for the built-in types :/ still, i think this reads so much nicer than seeing `value & 0xFFF` in the code that i think it's worth it (though definitely open to a better implementation method if there is one)

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
bgrozev: > thanks boris. but we don't leverage anything like picture id here

I think I misunderstood your question. 

> i don't get why we can make that guess if frame markings are supported but we don't know what they are

 I'm not sure that I understand it now either, so excuse me if that doesn't answer the question. If we can recognize frame boundaries (start and end), and we have a.getMaxSeen()=x and a.getEnd()=-1, we can deduce that x is not the end of a, therefore the end of a is after x.

bbaldino: ok, yeah agree with you there.  maybe we'll need to wait for @gpolitis to enlighten us on the rest :)

gpolitis: I'm sorry, I'm not following the discussion. But this code is not designed to leverage Picture IDs, it's designed to leverage frame markings which are available in the VPX payload descriptor and in the frame markings header extension. @bgrozev suggested the use of VPX picture IDs and I like the idea. 

>> why does this (frame markings support) allow us to infer anything extra from the sequence number differences?

Here's an example that hopefully clarifies things: First suppose that we get two frames that their diff is 3 and we haven't seen the start nor the end (n.b. we can always see the end by inspecting the RTP marker bit), e.g. we get this sequence [0, 3] with 0 from older frame and 3 from the newer frame.

Now, if both newer and older frames support frame markings, then we know there must be a start and an end in between 0 and 3, and since the only available slots are 1 and 2, 1 must be the end and 2 must be the start.

Unfortunately, if frame markings are not supported, then we can't deduce neither the start nor the end, because the end could be 1 or 2 (not 0 because it doesn't have the RTP marker bit set), and the start could be 2 or 3.

src/org/jitsi/util/RTPUtils.java
```diff
@@ -44,6 +44,7 @@ else if (diff > 1<<15)
      * (modulo 2^16).
      * @return result of the subtraction of one RTP sequence number from another
      * (modulo 2^16).
+     * FIXME(brian): why do we have this and sequenceNumberDiff?
```
gpolitis: sequenceNumberDiff takes into account the possibility of a sequence number wrap-around. substractNumber performs a substraction mod 2^16. substractNumber is poorly named :-)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/InterArrival.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.jitsi.impl.neomedia.rtp.remotebitrateestimator;
 
+import org.jitsi.impl.neomedia.rtp.TimestampUtils;
```
gpolitis: star import (you can setup your ide with this https://gist.github.com/gpolitis/19b75e4814638f8153ab2369a8599549)

test/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/TimestampUtilsTest.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.jitsi.impl.neomedia.rtp.remotebitrateestimator;
 
+import org.jitsi.impl.neomedia.rtp.TimestampUtils;
```
gpolitis: star import (you can setup your ide with this https://gist.github.com/gpolitis/19b75e4814638f8153ab2369a8599549)

src/org/jitsi/impl/neomedia/rtp/FrameDesc.java
```diff
@@ -183,6 +195,19 @@ public int getStart()
     }
 
     /**
+     * Returns whether or not the first sequence number of this frame
+     * is known (conclusively)
+     *
+     * @return true if we know the first sequence number of this frame, false
+     * otherwise
+     */
+    public boolean firstSequenceNumberKnown()
```
gpolitis: We can use this here as well: https://github.com/jitsi/jitsi-videobridge/blob/f3425d22f5b594891a5a47fc50cb40da25ce9efb/src/main/java/org/jitsi/videobridge/cc/SimulcastController.java#L738

src/org/jitsi/impl/neomedia/rtp/TimestampUtils.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+package org.jitsi.impl.neomedia.rtp;
+
+import org.jitsi.util.RTPUtils;
+
+/**
+ * Helper class to perform various timestamp manipulations and comparisons
+ * @author Ethan Lin
+ * //FIXME(brian): perhaps this should be moved alongside RTPUtils (or maybe its
+ * contents combined with RTPUtils?  This class deals with RTP timestamps)
```
gpolitis: I wouldn't mind that. A larger scale re-organization is going to happen at some point in the future because, unfortunately, there are methods for RTP/RTCP stuff manipulation all over the place, i.e. `RawPacket`, `RTCPHeaderUtils`, and whatnot.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
bbaldino: thanks george!  but how does the presence of frame markings alone (without actually looking at them) let us infer that?  is it because if packet 0 were marked as the end, we would have already known it at this point and it would've be set by now?  and, similarly, if packet 3 was the start we would have already known it at this point and it would've been set by now?

src/org/jitsi/util/RTPUtils.java
bbaldino: so are there still valid cases for both to exist?  should i rename `subtractNumber` to something else?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/InterArrival.java
bbaldino: fixed, thanks.  that xml file looks handy...where do i put it?

test/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/TimestampUtilsTest.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/rtp/TimestampUtils.java
bbaldino: if you have some thoughts there let me know, maybe i can clean some of that up next

bgrozev: We also have org.jitsi.util.TimeUtils. I didn't know about the existence of TimestampUtils when I added TimeUtils (d'oh!). The two can probably be merged.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/InterArrival.java
gpolitis: open IntelliJ IDEA Settings → Code Style → Java, click Manage, and import that XML file by simply clicking Import.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
gpolitis: > is it because if packet 0 were marked as the end, we would have already known it at this point and it would've be set by now? and, similarly, if packet 3 was the start we would have already known it at this point and it would've been set by now?

Yes, that's it!

gpolitis: That's done before the heuristics are applied.

bbaldino: 👍 cool, i get it now.  i think i'll add a comment to convey this.

src/org/jitsi/util/RTPUtils.java
gpolitis: Entirely your call :-) I can't come up with a preferred name.

src/org/jitsi/impl/neomedia/rtp/TimestampUtils.java
bbaldino: @bgrozev TimeUtils appears to be (mostly) NTP utils, which does feel like it's worth keeping separate (and maybe renaming to `NTPTimeUtils`?).  there is an `rtpDiff` method in there though that could be removed and usages replaced with the one in one of the other classes.

bbaldino:
couple more tweaks.  btw @gpolitis or @bgrozev please squash this before merging

bbaldino:
build is breaking because the bridge's libjitsi version hasn't been updated for george's breaking libjitsi changes.  pr for that here: https://github.com/jitsi/jitsi-videobridge/pull/506

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -249,124 +249,120 @@ public RTPEncodingDesc(
...
-                    // XXX(bgrozev): for VPX codecs with PictureID we could find
-                    // the start/end even with diff>2 (if PictureIDDiff == 1)
-
-                    // XXX(gp): we don't have the picture ID in FrameDesc and I
-                    // feel it doesn't belong there. We may need to subclass it
-                    // into VPXFrameDesc and H264FrameDesc and move the
-                    // heuristics logic in there.
-
-                    if (snDiff == 2)
+                    if (!olderFrame.lastSequenceNumberKnown())
```
gpolitis: can we keep the comment about using picture id as a further heuristic here?

test/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/TimestampUtilsTest.java
gpolitis: Is it?

PR #345, opened 2017-07-26
Adds a NullSrtpControl and minor fixes

bgrozev:


src/org/jitsi/impl/neomedia/RTPConnectorInputStream.java
```diff
@@ -743,9 +743,9 @@ private void runInReceiveThread()
 
         while (!closed)
         {
-            // http://code.google.com/p/android/issues/detail?id=24765
-            if (OSUtils.IS_ANDROID)
-                p.setLength(PACKET_RECEIVE_BUFFER_LENGTH);
+            // Reset the buffer, because the previous call to receive() might
+            // have bumped the offset or even changed the byte[].
+            p.setData(buffer, 0, buffer.length);
```
bbaldino: what's behind the change in logic here?

bgrozev: For Octo I override receive(), and I strip some headers from the packet by bumping the offset by 8 bytes. Without this reset the next call to receive uses offset=8, the underlying socket receives at that offset, then I strip 8 bytes and offset grow to 16, etc, etc.

bbaldino: ok, i think just more clarifying questions for my own sake:
1) the octo logic won't always kick in, will it?  does it make sense to have the reset (in the case of stripping the octo buffers) in the octo receive logic so its transparent elsewhere?
2) if i'm understanding what you said correctly (i don't think i am), how is the octo receive bumping the offset ahead different than the normal processing of a packet which reads the buffer?

bgrozev: > the octo logic won't always kick in, will it?
Correct.

> does it make sense to have the reset (in the case of stripping the octo buffers) in the octo receive logic so its transparent elsewhere?

I don't think this is possible, because the reset() needs to happen after the packet has been consumed by libjitsi. The flow is like this:
1. Libjitsi calls receive() with a buffer that it manages
2. The receive goes down to underlying sockets which copies data to the packet's buffer, at the packet's offset
3. Some Octo logic (part of the overriden DatagramSocket#receive()) reads the first 8 bytes and bumps the offset by 8
4. Libjitsi handles the received packet
5. Libjitsi calls receive() again

The reset happens between 4 and 5.



> if i'm understanding what you said correctly (i don't think i am), how is the octo receive bumping the offset ahead different than the normal processing of a packet which reads the buffer?

The Octo logic basically does p.setOffset(p.getOffset()+8). When libjitsi reads the packet it doesn't modify the offset, AFAIK.

bgrozev:
Jenkins please test this

PR #346, opened 2017-07-28
add rtpTimestampDiff and a helper function to RTPUtils

bbaldino:


src/org/jitsi/util/RTPUtils.java
```diff
@@ -217,4 +217,29 @@ else if (a > b)
...
+     */
+    public static long rtpTimestampDiff(long a, long b)
+    {
+        long diff = a - b;
+        if (diff < -(1L<<31))
+            diff+= 1L<<32;
+        else if (diff > 1L<<31)
+            diff-= 1L<<32;
+
+        return diff;
```
bgrozev: Why do we have two copies of this? Is TimeUtils#rtpDiff supposed to be removed?

bbaldino: actually i meant to mark it as deprecated but i was worried about removing it completely in case it's being used somewhere.  is it alright to just remove it?

bgrozev: I think so. Or it could call the new method.

bbaldino:
closing to use https://github.com/jitsi/libjitsi/pull/347 which has a branch name to match some other dependent changes

PR #347, opened 2017-07-28
add rtpTimestampDiff and a helper function to RTPUtils

bbaldino:


src/org/jitsi/util/TimeUtils.java
```diff
@@ -179,15 +179,11 @@ public static long getLsw(long ntpTime)
     /**
      * Returns the difference between two RTP timestamps.
      * @return the difference between two RTP timestamps.
+     * DEPRECATED: use RTPUtils.timeStampDiff
```
bgrozev: Typo. Perhaps we ca use "@deprecated use {@link RTPUtils#rtpTimestampDiff(long, long)}"?

Can we also update the now deprecated use of TimeUtils#rtpDiff within libjitsi?

bbaldino: good call, fixed.

PR #348, opened 2017-08-01
replace methods taking (buf, off, len) with ones taking RawPacket

bbaldino:


src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -4028,31 +4045,44 @@ public REDBlock getPrimaryREDBlock(ByteArrayBuffer baf)
...
+
+            return REDBlockIterator.getPrimaryBlock(pkt.getBuffer(), pkt.getOffset(), pkt.getLength());
         }
         else
         {
-            final int payloadOff = RawPacket.getPayloadOffset(buf, off, len),
-                payloadLen = RawPacket.getPayloadLength(buf, off, len, true);
-
-            return new REDBlock(buf, payloadOff, payloadLen, pktPT);
+            return new REDBlock(pkt.getBuffer(), pkt.getPayloadOffset(), pkt.getPayloadLength(), pktPT);
```
bgrozev: Line length

src/org/jitsi/impl/neomedia/codec/REDBlockIterator.java
```diff
@@ -16,6 +16,8 @@
 package org.jitsi.impl.neomedia.codec;
 
 import java.util.*;
+
+import org.jitsi.service.neomedia.*;
```
bgrozev: Why is this necessary?

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -413,16 +413,31 @@ RTPEncodingDesc findRTPEncodingDesc(long ssrc)
...
      * specified in the buffer passed in as an argument, or null.
+     * @Deprecated use findFrameDesc(long, long)
      */
+    @Deprecated
     public FrameDesc findFrameDesc(byte[] buf, int off, int len)
     {
-        RTPEncodingDesc rtpEncoding = findRTPEncodingDesc(buf, off, len);
-        if (rtpEncoding == null)
+        RawPacket pkt = new RawPacket(buf, off, len);
+        return findFrameDesc(pkt.getSSRCAsLong(), pkt.getTimestamp());
```
bgrozev: We could use the static versions of getSSRCAsLong and getTimestamp here

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -718,6 +718,23 @@ FrameDesc findFrameDesc(byte[] buf, int off, int len)
...
     /**
+     * Finds the {@link FrameDesc} that matches the RTP packet specified
+     * in the buffer passed in as an argument.
+     *
+     * @param timestamp the timestamp of the desired {@link FrameDesc}
+     *
+     * @return the {@link FrameDesc} that matches the RTP timestamp given,
+     * or null if there is no matching frame {@link FrameDesc}.
+     */
+    FrameDesc findFrameDesc(long timestamp)
```
bgrozev: Why do we keep the findFrameDesc(byte[] buf, int off, int len) method?

bgrozev:
Minor comments left inline. One more question: how do we handle the MediaStream interface?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/codec/REDBlockIterator.java
bbaldino: it's not, not sure how that got in there.  fixed.

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
bbaldino: i just hadn't replaced the existing chain, will remove it.

bbaldino:
it looks like jvb doesn't use the interface type anywhere, only MediaStreamImpl directly, that's why i didn't see any failures even though i didn't add it.  but added it to the interface now

PR #349, opened 2017-08-02
add 'isOlderSequenceNumberThan' and 'applySequenceNumberDelta' helper…

bbaldino:
… methods to RTPUtils.  rename 'sequenceNumberDiff' -> 'getSequenceNumberDelta'

i think  'delta' is clearer/more accurate than 'difference', since difference is typically thought of being a positive number ("the difference between 1 and 10 is 9.  the delta between 1 and 10 is -9").

the other helpers are things i've come across being done manually in jvb (jvb PR to come)

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
@@ -19,7 +19,6 @@ import net.sf.fmj.media.rtp.*;
 import org.jitsi.service.neomedia.*;
 import org.jitsi.util.*;
 
-import java.io.*;
 import java.util.*;
 
 /**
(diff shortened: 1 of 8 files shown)


bgrozev:
LGTM. Do you plan to submit one that updates the bridge? I think that'd be the easiest way to have this tested (not that I see anything that could brake).

bbaldino:
@bgrozev yeah good idea, opened https://github.com/jitsi/jitsi-videobridge/pull/512

jitsi-developers:
As a native English speaker, I don't find the difference between
"difference" and "delta" nearly as clear-cut as you suggest.  (I'd also
expect the delta between 1 and 10 to be 9, I could accept that the delta
between 10 and 1 is -9 though.  My point being that "the delta between a
and b" implies in my mind "a subtracted from b", not "a minus b".)

Martin

jitsi-developers:
... and I observe that the type used in C and C++ to return the (signed)
result of subtracting one pointer from another is "ptrdiff_t" - so there is
certainly precedence for using "difference" to mean "signed difference" in
computing.

Martin

PR #350, opened 2017-08-03
change the accept and willWriteXXX chains to use RawPacket instead of…

bbaldino:
… (buf, off, len)


bbaldino:
Jenkins, please test

PR #351, opened 2017-08-03
HQL-2762: The sender is not ramping up its sending bitrate

bbaldino:
we were not terminating REMBs sent from the audio ssrc correctly, so it's possible they could could be causing low bitrates from other participants.  this commit terminates them at the bridge so they don't make it all the way to far-end receivers.


bbaldino:
@gpolitis let me know if this is looking right

src/org/jitsi/impl/neomedia/rtcp/AudioRTCPTermination.java
```diff
@@ -0,0 +1,87 @@
...
+            super(RTCPPacketPredicate.INSTANCE);
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public RawPacket transform(RawPacket pkt)
+        {
+            // Kill the RRs that FMJ is sending.
```
bgrozev: Eeek

bbaldino: yeah i wasn't sure if we wanted to keep this for audio or not

bbaldino: ah nevermind, this comment is wrong now

bbaldino: i think i'll just put the logic from the other method directly in here

src/org/jitsi/impl/neomedia/rtcp/AudioRTCPTermination.java
```diff
@@ -0,0 +1,87 @@
...
+            return doTransform(pkt, true);
+        }
+
+        private RawPacket doTransform(RawPacket pkt, boolean send)
+        {
+            RTCPIterator it = new RTCPIterator(pkt);
```
bgrozev: I'm not going to protest too much, but perhaps we can abstract this kind of logic and avoid code duplication. E.g. we can have a method that removes RTCP packets given a filter.

PR #353, opened 2017-08-03
Fixes Absolute Send Time implementation

jcchuks:


jcchuks:
Please note: This PR now has  `RemoteBitrateEstimator.java`, `RemoteBitrateEstimatorAbsSendTime.java` and `RemoteBitrateEstimatorSingleStream.java` because of the dependency on RemoteBitrateEstimator interface. A new method has been declared the interface which is implemented by the AbsSendTime and SingleStream.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -168,16 +177,17 @@ private Cluster findBestProbe(List<Cluster> clusters)
...
         Cluster bestIt = new Cluster();
-        for (Cluster cluster : clusters) {
+        for (Cluster cluster : clusters)
+        {
             if (cluster.sendMeanMs == 0 || cluster.recvMeanMs == 0)
                 continue;
             if (cluster.numAboveMinDelta > cluster.count / 2 &&
-                    (cluster.recvMeanMs - cluster.sendMeanMs <= 2.0f &&
-                            cluster.sendMeanMs - cluster.recvMeanMs <= 5.0f))
+                (cluster.recvMeanMs - cluster.sendMeanMs <= 2.0f &&
```
MartinBonner: I would find this easier to read as:

```
                (-2.0f <= cluster.sendMeanMs - cluster.recvMeanMs &&
                          cluster.sendMeanMs - cluster.recvMeanMs <= 5.0f))
```

Because of the  `cluster.` and `MeanMs` prefix and suffix, it is really hard to see that the first difference is the negation of the second.  By multiplying both sides by -1 (and hence inverting the sense of the comparison), it a) becomes much easier to see the symmetry, b) becomes my preferred "in range check".

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -21,74 +21,83 @@
...
         this.estimator = new OveruseEstimator(new OverUseDetectorOptions());
         this.detector = new OveruseDetector(new OverUseDetectorOptions());
-        this.incomingBitrate = new RateStatistics(kBitrateWindowMs,8000);
+        this.incomingBitrate = new RateStatistics(kBitrateWindowMs, 8000F);
         this.incomingBitrateInitialized = false;
         this.totalProbesReceived = 0;
         this.firstPacketTimeMs = -1;
         this.lastUpdateMs = -1;
-        logger.info("; RemoteBitrateEstimatorAbsSendTime: Instantiating.");
+        logger.info("; RemoteBitrateEstimatorAbsSendTime: Calling.");
```
bgrozev: Can we change that to debug level please (with an if(logger.isDebugEnabled()) check).

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -579,54 +576,18 @@ public void setMinBitrate(int minBitrateBps)
...
+     * {@code sendSideEstimate} is the wrapper for bitrate estimation when
+     * the bridge is the sender.
+     * @param ssrc is the SSRC associated with the RTP packet send by the bridge
+     * @param arrivalTimeMs is time the packet reached the client
+     * @param sendTimeMs is time the packet left the bridge
+     * @param packetSize is the size of the packet.
      */
     @Override
-    public void run()
+    public void sendSideEstimate(long ssrc, long arrivalTimeMs, long sendTimeMs, long packetSize)
```
bgrozev: Why do we need to add a new method if it just calls incomingPacketInfo() with the arguments rearranged? For send-side bwe we know we will only be using RemoteBitrateEstimatorAbsSendTime so there's no need to pollute the RemoteBitrateEstimator with extra methods.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSingleStream.java
```diff
@@ -305,11 +303,26 @@ public void setMinBitrate(int minBitrateBps)
...
+
+    }
+
+    /**
      * Triggers a new estimate calculation.
      *
      * @param nowMs
      */
-    private void updateEstimate(long nowMs)
+    private void updateEstimate(long nowMs, long ssrc)
```
bgrozev: We don't want this code used for trace logging in the master repo.

bgrozev:
This needs to be split it two PRs -- one for the fixes to RBEAbsSendTime and one for the send-side bwe work.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Hi Martin,  Yes, that is true. Thanks for the suggestion.  Will update it.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Okay

jcchuks:
@bgrozev, Alright. Can I still include all files affected by the implementation of the send-side bwe in one PR. (TransportCCEngine inclusive)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Alright

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -27,123 +27,132 @@
...
-        for(Map.Entry<K,V> entry : _map.entrySet())
+        for (Map.Entry<K, V> entry : _map.entrySet())
         {
             keys.add(entry.getKey());
         }
         return keys;
     }
 
-    private long ConvertMsTo24Bits(long timeMs)
+    public long ConvertMsTo24Bits(long timeMs)
```
bbaldino: does this operate on an rtp timestamp?  we have some helpers in RTPUtils for this sort of thing, could we add this there?

bbaldino: even if not, if this is useful in more places (assuming so, since it's public) it'd be nice to have this as a static helper function in a utils file

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -27,123 +27,132 @@
...
-        if(clusterAggregate.count == 0)
+        if (clusterAggregate.count == 0)
         {
             return true;
         }
         double clusterMean = clusterAggregate.sendMeanMs /
-                (double)clusterAggregate.count;
-        return  Math.abs((double)sendDeltaMs - clusterMean) < 2.5f;
+            (double) clusterAggregate.count;
+        return Math.abs((double) sendDeltaMs - clusterMean) < 2.5f;
```
bbaldino: could we get a named constant for `2.5f`?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -27,123 +27,132 @@
...
     }
 
     private void computeClusters(List<Cluster> clusters)
     {
         Cluster current = new Cluster();
-        long prevSendTime =  -1;
+        long prevSendTime = -1;
         long prevRecvTime = -1;
-        for(Probe probe : probes_)
+        for (Probe probe : probes_)
```
bbaldino: what is with the naming of `probes_`? i've not seen that convention anywhere

bbaldino: i just realized perhaps all this is to be more consistent with the original cpp code, if so then can ignore all these

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -168,16 +177,17 @@ private Cluster findBestProbe(List<Cluster> clusters)
...
-        for (Cluster cluster : clusters) {
+        for (Cluster cluster : clusters)
+        {
             if (cluster.sendMeanMs == 0 || cluster.recvMeanMs == 0)
                 continue;
             if (cluster.numAboveMinDelta > cluster.count / 2 &&
-                    (cluster.recvMeanMs - cluster.sendMeanMs <= 2.0f &&
-                            cluster.sendMeanMs - cluster.recvMeanMs <= 5.0f))
+                (-2.0f <= cluster.sendMeanMs - cluster.recvMeanMs &&
+                    cluster.sendMeanMs - cluster.recvMeanMs <= 5.0f))
```
bbaldino: constants for `-2.0f`, `5.0f`?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -264,23 +276,23 @@ private boolean isBitrateImproving(int newBitrateBps)
...
-
         incomingPacketInfo(System.currentTimeMillis(), absoluteSendTimeEngine
                 .getAbsSendTime(packet), packet.getPayloadLength(),
-                packet.getSSRCAsLong());
+            packet.getSSRCAsLong());
         return packet;
     }
 
-    private void incomingPacketInfo(
+    public void incomingPacketInfo(
```
bbaldino: this seems like an odd name.  seems like it should be `processIncomingPacketInfo` or `onIncomingPacketInfo` or something

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -293,56 +305,64 @@ private void incomingPacketInfo(
...
+
         if (incomingBitrate_ != 0)
         {
             incomingBitrateInitialized = true;
-        } else if (incomingBitrateInitialized)
+        }
+        else if (incomingBitrateInitialized)
         {
             // Incoming bitrate had a previous valid value, but now not
             // enough data point are left within the current window.
```
bbaldino: where is the logic that tests that there is no longer enough data points left in the current window? is there a test for that somewhere?  or maybe i'm mis-reading this.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -353,33 +373,33 @@ private void incomingPacketInfo(
 
             long[] deltas = new long[]{tsDelta, tDelta, sizeDelta};
```
bbaldino: what's the point of putting these values into this array?

bbaldino: oops, missed where the array gets passed to a function.  nevermind.

bbaldino: is there any reason not to use them by their original names in the code below though? reading `tsDelta` is nicer than `deltas[0]`

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -484,13 +526,15 @@ public PacketTransformer getRTCPTransformer() {
...
+    public Collection<Integer> getSsrcs()
+    {
 
         synchronized (critSect)
         {
             Collection<Integer> ssrcs
-                    = new ArrayList<>();
-            for(Long ssrcValue : ssrcs_.keySet()){
+                = new ArrayList<>();
+            for (Long ssrcValue : ssrcs_.keySet())
```
bbaldino: naming of `ssrcs_`?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: I believe after passing arrays as params to the computeDelta function  (the computeDelta function modifies that array), the only way to retrieve the modified values of the arrays is to get them by referencing the array i.e `deltas[0]` since arrays are objects & on the heap. Using `tsDelta` directly will give us the value of `tsDelta `initially set since `tsDelta `is of primitive type on the stack, and not scoped within computeDelta function. In the cpp code, the values where passed & retrieved by reference i.e `&tsDelta`. Let me know if there is any other way. Thanks.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Sorry, I didn't get this.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Thank @bbaldino for the comment. I think if we read it as when `incomingBitrate_ == 0` and `incomingBitrateInitialized == true`, then I think it makes sense to say the 

> Incoming bitrate had a previous valid value, but now not
    enough data point are left within the current window

Since this would mean that `incomingBitrate_ ` had a previous valid value thus making `incomingBitrateInitialized `true but now equals `zero ` which could mean that there aren't enough data points.
I think this translates to the logic flow for setting `incomingBitrateInitialized` and checking when `incomingBitrate_ ` is zero. Correct me if am wrong. Thanks.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Alright, will update this

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Please can you suggest name for the constants. ?
Thanks.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Okay :)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
jcchuks: Alright, makes sense to transfer this to `RTPUtils`.,  For `kAbsSendTimeFraction` constant,
I will pass it as a parameter to the new `convertMsTo24Bits` in `RTPUtils`. Let me know if you have a preferred alternative to handling the constant `kAbsSendTimeFraction`.

gpolitis:
Superseded by #361.

PR #355, opened 2017-08-07
Logs the measured/estimated packet delay gradient.

gpolitis:


src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSingleStream.java
```diff
@@ -65,7 +64,7 @@
 
     private long lastProcessTime = -1L;
 
-    private final RemoteBitrateObserver observer;
+    private final VideoMediaStreamImpl stream;
```
bgrozev: I would prefer to keep the code as close to the original as possible -- keep the observer as "observer", and have VideoMediaStreamImpl implement the interface. What do you think? We can still call hashCode on it, which seems like the only use so far (but perhaps you have something else in mind for the future).

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSingleStream.java
```diff
@@ -35,7 +35,6 @@
 public class RemoteBitrateEstimatorSingleStream
     extends SinglePacketTransformerAdapter
     implements CallStatsObserver,
-               RecurringRunnable,
```
bgrozev: Why are we doing this (given that it's a single executor that run them all, so we're not saving threads)? Do you think this (executing only when packets arrive) will affect the performance of the RBE?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSingleStream.java
gpolitis: Will do as you suggest. Just a remark, the observer doesn't do anything actually, i.e. we're not acting on any observations, so we may want to completely remove this in the future.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSingleStream.java
gpolitis: I see this as code cleanup because it makes sense to update the estimate when you have data (and not update it when you don't).

This won't run more often than before (at most every 500ms). It may run less often, if we don't receive any RTP packets or if we receive RTP packets less frequently. There are no expensive operations in the updateEstimate method, so I doubt this will have any noticeable impact on performance (either positive or negative), but I don't have any data to support this claim.

bgrozev: OK, LGTM

PR #357, opened 2017-08-15
adds convertMsTo24Bits to RTPUtils

jcchuks:
moves `convertMsTo24Bits` method from RemoteBitrateEstimatorAbsSendTime to RTPUtils.java

### DIFF
--- a/src/org/jitsi/util/RTPUtils.java
+++ b/src/org/jitsi/util/RTPUtils.java
@@ -24,10 +24,14 @@ import java.util.*;
 public class RTPUtils
 {
     /**
-     * Returns the difference between two RTP sequence numbers (modulo 2^16).
-     * @return the difference between two RTP sequence numbers (modulo 2^16).
+     * Returns the delta between two RTP sequence numbers, taking into account
+     * rollover.  This will return the 'shortest' delta between the two
+     * sequence numbers in the form of the number you'd add to b to get a. e.g.:
+     * getSequenceNumberDelta(1, 10) -> -9 (10 + -9 = 1)
+     * getSequenceNumberDelta(1, 65530) -> 7 (65530 + 7 = 1)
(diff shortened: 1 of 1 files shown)


bgrozev:
In general, I don't mind moving this method to the utils class. However:
1. The method still exists in RemoteBitrateEstimatorAbsSendTime
2. It is not used anywhere. I see that you use it in TransportCCEngine in your branch, but I this is incorrect -- we shouldn't be saving the RTP timestamps of outgoing packets, but the time at which the packets we actually send.

jcchuks:
I removed the method in my last commit on RBEAbsSendTime.

> It is not used anywhere. I see that you use it in TransportCCEngine in your branch, but I this is incorrect -- we shouldn't be saving the RTP timestamps of outgoing packets, but the time at which the packets we actually send.

Okay, makes sense. I thought about using `System.currentTimeToMillis()` in place of the RTP timestamp, but it doesn't look like it too.  My thought is that this timestamp should be obtained using `System.currentTimeToMillis()` somewhere after the bandwidth controller has scheduled and sent the packet. Please how do I retrieve the correct timestamp?

bgrozev:
There aren't any components on the bridge that would delay a packet. Some might drop packets, or make some modifications, but that's OK for your use. The local system time saved at the time the packet passed through the transport-cc engine should work.

jcchuks:
Okay, nice 


Sent from Yahoo Mail for iPhone


On Wednesday, August 16, 2017, 11:30 AM, bgrozev <notifications@github.com> wrote:


There aren't any components on the bridge that would delay a packet. Some might drop packets, or make some modifications, but that's OK for your use. The local system time saved at the time the packet passed through the transport-cc engine should work.

gpolitis:
Superseded by #361.

PR #359, opened 2017-08-22
update minSeen/maxSeen when setting start/end in FrameDesc.

bbaldino:
also add a helper func for getting all the encodings for a track via the FrameDesc.

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/FrameDesc.java
+++ b/src/org/jitsi/impl/neomedia/rtp/FrameDesc.java
@@ -187,10 +187,6 @@ public class FrameDesc
     void setEnd(int end)
     {
         this.end = end;
-        if (this.end > this.maxSeen)
-        {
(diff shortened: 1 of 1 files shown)


src/org/jitsi/impl/neomedia/rtp/FrameDesc.java
```diff
@@ -182,6 +187,10 @@ public boolean lastSequenceNumberKnown()
     void setEnd(int end)
     {
         this.end = end;
+        if (this.end > this.maxSeen)
```
gpolitis: Here we need to take sequence number wrap around into account. Alternatively, what we could do is unconditionally set the `maxSeen`, and prevent the `maxSeen` from being set if the `end` has been set.

However, I don't like adding this kind of logic in the setters. Maybe it'd be better to have a getMaxOrEnd() method.

PR #360, opened 2017-08-24
change sr packet model and parsing over to new libjitsi classes

bbaldino:


bbaldino:
@bgrozev i don't think this is ready to merge yet, i have a couple questions and i think it's likely we'll want to tweak a couple things, but this should give a good idea of what the process of moving these over will look like, so let me know what you think of the general approach.

src/org/jitsi/impl/neomedia/rtcp/NewRTCPPacketParser.java
```diff
@@ -0,0 +1,66 @@
...
+     * instances for logging output.
+     */
+    private static final Logger logger
+        = Logger.getLogger(NewRTCPPacketParser.class);
+
+    public static NewRTCPPacket[] parse(RawPacket packet)
+    {
+        // PR-NOTE(brian): should we do some check/verification that this is, in fact,
+        // a compound rtcp packet?  Or can we expect to only be called with
+        // compound packets?
```
bgrozev: Unless I'm missing something, non-compound packets can be treated as a special case of compound packets (with only one packet)

src/org/jitsi/impl/neomedia/rtcp/NewRTCPPacketParser.java
```diff
@@ -0,0 +1,66 @@
...
+                        }
+                        // We remove here because, for now, this parser is only
+                        // parsing a subset of types, so this compound packet
+                        // will later be passed to the old parser as well.
+                        // We'll remove the ones that we handle here and combine
+                        // all the parsed packets at the call site.  In an
+                        // effort to not mask problems with the new code, we'll
+                        // remove it even if we failed to parse it for some reason
+                        // (since, when detecting an sr packet, we should have
+                        // parsed it successfully)
```
bgrozev: I'm not sure this approach will work well. If on the first stage we are only replacing the parsing code in StatisticsEngine, then the SR will still need to reach FMJ like it does now.

src/org/jitsi/impl/neomedia/transform/rtcp/StatisticsEngine.java
```diff
@@ -977,10 +983,10 @@ private void updateReceivedMediaStreamStats(
                 break;
 
             case RTCPPacket.SR:
-                if (rtcp instanceof RTCPSRPacket)
-                {
-                    streamStats.srReceived((RTCPSRPacket) rtcp);
-                }
+                streamStats.srReceived((NewRTCPSRPacket) rtcp);
+                //FIXME(brian): this break wasn't here before, was that
+                //intentional?
```
bgrozev: Looks intentional, given that parseRTCPReport supports both SR and RR

src/org/jitsi/impl/neomedia/rtcp/NewRTCPPacketParser.java
bbaldino: but the iterator will only work on compound packets, is that right?

bgrozev:
@bbaldino does this need a review?

bbaldino:
closing this, going to go about this differently.

PR #361, opened 2017-08-28
feat: Implements send-side bandwidth estimations.

gpolitis:
Depends on https://github.com/jitsi/jicofo/pull/216
Provides https://github.com/jitsi/jitsi-videobridge/pull/529

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -344,7 +344,7 @@ public class MediaStreamImpl
      * The ID of the frame markings RTP header extension. We use this field as
      * a cache, in order to not access {@link #activeRTPExtensions} every time.
      */
-    private byte frameMarkingsExtensionId = -1;
+    private int frameMarkingsExtensionId = -1;
 
     /**
      * The {@link TransportCCEngine} instance, if any, for this
@@ -668,7 +668,7 @@ public class MediaStreamImpl
         boolean active
             = !MediaDirection.INACTIVE.equals(rtpExtension.getDirection());
 
-        byte effectiveId = active ? extensionID : -1;
+        int effectiveId = active ? RTPUtils.as16Bits(extensionID) : -1;
 
         String uri = rtpExtension.getURI().toString();
         if (RTPExtension.ABS_SEND_TIME_URN.equals(uri))
@@ -684,7 +684,7 @@ public class MediaStreamImpl
             if (remoteBitrateEstimatorWrapper != null)
             {
                 remoteBitrateEstimatorWrapper
-                    .setExtensionID(RTPUtils.as16Bits(effectiveId));
+                    .setExtensionID(effectiveId);
             }
         }
         else if (RTPExtension.FRAME_MARKING_URN.equals(uri))
@@ -3670,7 +3670,7 @@ public class MediaStreamImpl
         if (frameMarkingsExtensionId != -1)
         {
             RawPacket.HeaderExtension fmhe
-                = pkt.getHeaderExtension(frameMarkingsExtensionId);
+                = pkt.getHeaderExtension((byte) frameMarkingsExtensionId);
 
             if (fmhe != null)
             {
@@ -3756,7 +3756,7 @@ public class MediaStreamImpl
             RawPacket pkt = new RawPacket(buf, off, len);
 	    String encoding = getFormat(pkt.getPayloadType()).getEncoding();
             RawPacket.HeaderExtension fmhe
-                = pkt.getHeaderExtension(frameMarkingsExtensionId);
+                = pkt.getHeaderExtension((byte) frameMarkingsExtensionId);
             if (fmhe != null)
             {
                 return FrameMarkingHeaderExtension.getSpatialID(fmhe,encoding);
@@ -3821,7 +3821,7 @@ public class MediaStreamImpl
         }
         else
         {
-            return pkt.getHeaderExtension(frameMarkingsExtensionId) != null;
+            return pkt.getHeaderExtension((byte) frameMarkingsExtensionId) != null;
         }
     }
 
@@ -3847,7 +3847,7 @@ public class MediaStreamImpl
         if (frameMarkingsExtensionId != -1)
         {
             RawPacket.HeaderExtension fmhe
-                = pkt.getHeaderExtension(frameMarkingsExtensionId);
+                = pkt.getHeaderExtension((byte) frameMarkingsExtensionId);
             if (fmhe != null)
             {
                 return FrameMarkingHeaderExtension.isStartOfFrame(fmhe);
@@ -3907,7 +3907,7 @@ public class MediaStreamImpl
         if (frameMarkingsExtensionId != -1)
         {
             RawPacket.HeaderExtension fmhe
-                = pkt.getHeaderExtension(frameMarkingsExtensionId);
+                = pkt.getHeaderExtension((byte) frameMarkingsExtensionId);
             if (fmhe != null)
             {
                 return FrameMarkingHeaderExtension.isEndOfFrame(fmhe);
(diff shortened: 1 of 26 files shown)


src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -676,6 +677,15 @@ private void enableRTPExtension(byte extensionID, RTPExtension rtpExtension)
...
                 absSendTimeEngine.setExtensionID(effectiveId);
             }
+
+            RemoteBitrateEstimatorWrapper remoteBitrateEstimatorWrapper
+                = getRemoteBitrateEstimator();
+
+            if (remoteBitrateEstimatorWrapper != null)
+            {
+                remoteBitrateEstimatorWrapper
+                    .setExtensionID(effectiveId & 0xFFFF);
```
bbaldino: could this use `RTPUtils.as16Bits()`?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -4130,7 +4140,7 @@ protected PaddingTermination getPaddingTermination()
      * @return the <tt>RemoteBitrateEstimator</tt> of this
      * <tt>VideoMediaStream</tt> if any; otherwise, <tt>null</tt>
      */
-    public RemoteBitrateEstimator getRemoteBitrateEstimator()
+    public RemoteBitrateEstimatorWrapper getRemoteBitrateEstimator()
```
bbaldino: should the name of this method change to `getRemoteBitrateEstimatorWrapper() to match`?  btw, i don't care too much if we have a separate type for the wrapper or if we continue to call it `RemoteBitrateEstimator` and just have it contain an inner delegate that does the real estimation.

src/org/jitsi/impl/neomedia/VideoMediaStreamImpl.java
```diff
@@ -448,8 +448,8 @@ else if(closestAspect == null)
      * The <tt>RemoteBitrateEstimator</tt> which computes bitrate estimates for
      * the incoming RTP streams.
      */
-    private final RemoteBitrateEstimator remoteBitrateEstimator
-        = new RemoteBitrateEstimatorSingleStream(
+    private final RemoteBitrateEstimatorWrapper remoteBitrateEstimator
+        = new RemoteBitrateEstimatorWrapper(
```
bbaldino: i'm confused that we have a different type here (`RemoteBitrateEstimatorWrapper` instead of `RemoteBitrateEstimator`) given that we initialize the 'wrapper' the same way as we did the estimator itself, is the 'wrapper' effectively an estimator?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -263,15 +258,20 @@ private boolean isBitrateImproving(int newBitrateBps)
...
+            long ast =
+                AbsSendTimeEngine.getAbsSendTime(packet, (byte) extensionID);
 
-        incomingPacketInfo(System.currentTimeMillis(), absoluteSendTimeEngine
-                .getAbsSendTime(packet), packet.getPayloadLength(),
-                packet.getSSRCAsLong());
-        return packet;
+            if (ast != -1)
+            {
+                incomingPacketInfo(System.currentTimeMillis(), ast,
```
bbaldino: the name of this method seems odd to me, can we change it to a verb?  `incomingPacketInfo` sounds like a type instead of a method (`processIncomingPacketInfo`?)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
```diff
@@ -0,0 +1,221 @@
...
+     * Determines whether or not to activate the abs-send-time remote bitrate
+     * estimator.
+     */
+    private static final boolean ENABLE_AST_RBE =
+        cfg != null ? cfg.getBoolean(ENABLE_AST_RBE_PNAME,
+            ENABLE_AST_RBE_DEFAULT) : ENABLE_AST_RBE_DEFAULT;
+
+    /**
+     *
+     */
```
bbaldino: some missing javadocs in here

src/org/jitsi/impl/neomedia/transform/AbsSendTimeEngine.java
```diff
@@ -163,19 +143,23 @@ public byte getAbsSendTimeExtensionID(){
      * @param pkt is a RawPacket
      * @return
      */
-    public long getAbsSendTime(RawPacket pkt)
+    public static long getAbsSendTime(RawPacket pkt, byte extensionID)
```
bbaldino: yay for static methods where possible :)

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -676,6 +677,15 @@ private void enableRTPExtension(byte extensionID, RTPExtension rtpExtension)
             {
                 absSendTimeEngine.setExtensionID(effectiveId);
             }
+
+            RemoteBitrateEstimatorWrapper remoteBitrateEstimatorWrapper
```
bbaldino: we don't actually need to declare this as a `RemoteBitrateEstimatorWrapper` here, right?  we probably only need the `RemoteBitrateEstimator` type?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -1103,7 +1113,7 @@ private TransformEngineChain createTransformEngineChain()
 
         // TODO RTCP termination should end up here.
 
-        RemoteBitrateEstimator
+        RemoteBitrateEstimatorWrapper
```
bbaldino: again, can we get away with just the `RemoteBitrateEstimator` type?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
bbaldino: can this just return the `RemoteBitrateEstimator` type?

src/org/jitsi/impl/neomedia/VideoMediaStreamImpl.java
```diff
@@ -928,7 +928,7 @@ public QualityControl getQualityControl()
      * {@inheritDoc}
      */
     @Override
-    public RemoteBitrateEstimator getRemoteBitrateEstimator()
+    public RemoteBitrateEstimatorWrapper getRemoteBitrateEstimator()
```
bbaldino: can this just return the `RemoteBitrateEstimator` type?

gpolitis:
> with regards to the RemoteBitrateEstimatorWrapper type: i'm not sure why we have a type which explicitly refers to it as a wrapper, but in some places we obfuscate it with calls like getBitrateEstimator (as opposed to getBitrateEstimatorWrapper). if the caller only cares about the fact that it's getting a bitrate estimator, maybe we should remove the 'wrapper' part and just refer to it as an estimator, while documenting in the class itself that it delegates all the work to a real implementation?

The wrapper also implements the `TransformEngine` interface. I could either rename `MediaStreamImpl.getRemoteBitrateEstimator()` to `MediaStreamImpl.getRemoteBitrateEstimatorWrapper()` OR introduce a 2nd method called `MediaStreamImpl.getRemoteBitrateEstimatorTransformEngine()` and both methods would return the wrapper.

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
gpolitis: The `RemoteBitrateEstimator` interface doesn't have a `setExtension` method (nor should it).

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
gpolitis: The `RemoteBitrateEstimator` is not a `TransformEngine` (nor should it).

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
gpolitis: No, we can't return a `RemoteBitrateEstimator` because of the above two comments; but I can rename the method if this is your suggestion.

PR #363, opened 2017-09-19
Minor cleanup and refactoring

bgrozev:
Removes some deprecated methods from MediaStreamImpl (just the Impl)

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -3788,33 +3788,30 @@ public class MediaStreamImpl
      * Utility method that determines the spatial layer index (SID) of an RTP
      * packet.
      *
-     * @param buf the buffer that holds the RTP payload.
-     * @param off the offset in the buff where the RTP payload is found.
-     * @param len then length of the RTP payload in the buffer.
+     * @param pkt the RTP packet.
      *
      * @return the SID of the packet, -1 otherwise.
      *
      * FIXME(gp) conceptually this belongs to the {@link VideoMediaStreamImpl},
      * but I don't want to be obliged to cast to use this method.
      */
-    public int getSpatialID(byte[] buf, int off, int len)
+    public int getSpatialID(RawPacket pkt)
     {
-        if (frameMarkingsExtensionId != -1) 
-	{
-            RawPacket pkt = new RawPacket(buf, off, len);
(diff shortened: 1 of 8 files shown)


src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -3760,56 +3765,33 @@ else if (redBlock.getPayloadType() == vp9PT)
...
-     * @param len then length of the RTP payload in the buffer.
+     * @param pkt the RTP packet.
      *
      * @return the SID of the packet, -1 otherwise.
      *
      * FIXME(gp) conceptually this belongs to the {@link VideoMediaStreamImpl},
      * but I don't want to be obliged to cast to use this method.
      */
-    public int getSpatialID(byte[] buf, int off, int len)
+    public int getSpatialID(RawPacket pkt)
```
bbaldino: 🎉

src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -352,23 +352,20 @@ void update(
...
+     * @param pkt the packet to match.
      * @return the {@link RTPEncodingDesc} that corresponds to the packet that is
      * specified in the buffer passed in as an argument, or null.
      */
-    RTPEncodingDesc findRTPEncodingDesc(byte[] buf, int off, int len)
+    RTPEncodingDesc findRTPEncodingDesc(RawPacket pkt)
     {
-        if (buf == null || buf.length < off + len
-            || ArrayUtils.isNullOrEmpty(rtpEncodings))
+        if (ArrayUtils.isNullOrEmpty(rtpEncodings))
```
bbaldino: should we add a check for `!pkt.isInvalid()`?

bgrozev: Sure. I would add it either on top of the chain (so that it triggers early) or at the bottom (so that it is close to the code which actually needs it). Or maybe even both, but I don't like it here in the middle. Any preference?

bbaldino: at the top of the chain sounds good.  maybe tweak the javadoc to show that it assumes/requires the `RawPacket` is non-null & valid in the places down the chain?

PR #364, opened 2017-09-22
fix: Fixes inter-arrival deltas computation (again).

gpolitis:


bgrozev:
It's hard to distinguish between refactoring and the actual fix. All the refactoring looks good to me, except that I would keep the unit tests and port them to use the new method.

Are the 3 places where "TimestampUtils.subtractAsUnsignedInt32" is replaced with "RTPUtils.rtpTimestampDiff" intentional?

PR #366, opened 2017-09-25
bring ByteBufferOutputStream from fmj

bbaldino:
this is part of the effort to remove the fmj dependency

### DIFF
--- a/src/org/jitsi/util/ByteBufferOutputStream.java
+++ b/src/org/jitsi/util/ByteBufferOutputStream.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright @ 2017 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.util;
+
+import java.io.*;
+
+/**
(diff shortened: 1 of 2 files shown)


src/org/jitsi/util/ByteBufferOutputStream.java
```diff
@@ -0,0 +1,113 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void write(int b)
+        throws IOException
+    {
+        if (index >= endIndex)
+        {
+            throw new IOException(
+                "This " + net.sf.fmj.utility.ByteBufferOutputStream.class.getName()
```
ibauersachs: This isn't an FMJ class anymore.

ibauersachs:
> this is part of the effort to remove the fmj dependency

Maybe I missed something because I'm not always in The Call. Could you please clarify that? AFAIK FMJ is the core port for rendering all audio/video (in desktop applications).

src/org/jitsi/util/ByteBufferOutputStream.java
bbaldino: ah good catch, will get this fixed

bbaldino:
> > this is part of the effort to remove the fmj dependency
> 
> Maybe I missed something because I'm not always in The Call. Could you please clarify that? AFAIK FMJ is the core port for rendering all audio/video (in desktop applications).

We no longer need that functionality in the bridge and it's making some other work difficult, so the idea is to implement the rtp stack logic that's needed by the bridge in libjitsi so we can remove the dependency on fmj there.  from talking with emil, i think we should then be able to change fmj to rely on libjitsi for the rtp stack so we don't have any duplication.  other components which need fmj can keep using it for the encoding/decode logic.

ibauersachs:
Fine, it might even improve some broken parts. But please make sure that Jitsi Desktop still works. I don't expect you guys to actively support it, but please don't break it either. I can't afford maintaining a fork of libjitsi.

bgrozev:
Jigasi and jitsi-hammer depend on fmj and there are no plans to remove that dependency, so I don't think there's any risk for the desktop client.

PR #369, opened 2017-09-29
Sets vp8 as default codec.

damencho:


bbaldino:
what's the plan here?  if anything we shouldn't comment out h264, just move it later in the list.

PR #370, opened 2017-10-04
DO NOT MERGE.  port over subset of the rtp-stack changes from fmj

bbaldino:
focusing on `StatisticsEngine`, port over any classes from fmj into libjitsi.

a (small) jvb change was necessary to accommodate this as well which can be found here: https://github.com/bbaldino/jitsi-videobridge/tree/fmj_1

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamStatsImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamStatsImpl.java
@@ -26,10 +26,9 @@ import javax.media.format.*;
 import javax.media.protocol.*;
 import javax.media.rtp.*;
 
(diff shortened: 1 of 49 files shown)


ibauersachs:
You cannot copy code from fmj into libjitsi like that: fmj is LGPL licensed, so the copied files need to keep that license, which in turn would make libjitsi a mixed-license project.
Since that's apparantly  not an option, jitsi-lgpl-dependencies exists.

It's okay of course if you have a relicensing permission from all previous authors, but then please add a clear note about that.

bbaldino:
hey @ibauersachs, there's no plan to commit this code.  this is just an intermediary step in the process to try and implement this in phases rather than just ripping it all out at once.

PR #378, opened 2017-11-04
fixed readInt16AsInt implementation

sidorovis:


bgrozev:
Thanks for the fix! Please let us know if you have already signed the CLA

sidorovis:
I believe so.

Also I'm thinkging about refactoring transport-cc's PacketMap to some kind of List / Array struct. Because in case when we receive more then 32768 feedbacks from client side it will work incorrectly.
And it would be much faster.

bgrozev:
Do you mean RTCPTCCPacket.PacketMap? If so, it is meant to store the packets for/from a single RTCP packet, so in most cases <50 entries. If those entries span a range of more than 32768 that would indeed be a problem, but we don't expect this to happen because at the highest packet rates we see it takes about 2 minutes.

About performance: in the vast majority of cases, the map will be filled sequentially. Assuming that TreeMap's insert is log(N) (I haven't checked if that's true), and that it stays with <50 elements, it shouldn't cause any problems.

To be cleanr we'd be happy to review and accept contributions, I'm just not sure that this is something worth spending time on.

PR #383, opened 2017-11-14
WIP: Implements send-side bandwidth estimations.

gpolitis:
- Propagates RTT to the send-side delay-based controller.
- Partially reverts most of the RTCP TCC parsing changes that were
  introduced in e019b4217daa3967efa457c299f76c1f92e012f3.
- Uses correct SSRCs in the outgoing RTCP TCC packets.
- Fixes a synchronization which could lead to undefined behavior of the
  Java Map.

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
@@ -167,7 +167,7 @@ public class RTCPReceiverFeedbackTermination
             return -1;
         }
 
-        return stream.getStreamRTPManager().getLocalSSRC();
+        return streamRTPManager.getLocalSSRC();
     }
 
 
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
@@ -415,6 +415,7 @@ public class RTCPReceiverFeedbackTermination
                         || RTCPTCCPacket.isTCCPacket(baf))
                 {
                     it.remove();
+                    continue;
                 }
 
                 if (!send)
@@ -431,6 +432,7 @@ public class RTCPReceiverFeedbackTermination
                             .requestKeyframe(source);
 
                         it.remove();
+                        continue;
                     }
                 }
             }
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
@@ -410,7 +410,9 @@ public class RTCPReceiverFeedbackTermination
             {
                 ByteArrayBuffer baf = it.next();
                 int pt = RTCPHeaderUtils.getPacketType(baf);
-                if (pt == RTCPRRPacket.RR || RTCPREMBPacket.isREMBPacket(baf))
+                if (pt == RTCPRRPacket.RR
+                        || RTCPREMBPacket.isREMBPacket(baf)
+                        || RTCPTCCPacket.isTCCPacket(baf))
                 {
                     it.remove();
                 }
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
@@ -90,6 +90,14 @@ public class RTCPTCCPacket
         return getPacketsFci(getFCI(baf));
     }
 
+    /**
+     * @return the reference time of the FCI buffer of an RTCP TCC packet.
+     *
+     * The format is 32 bits with 250µs resolution. Note that the format in the
+     * transport-wide cc draft is 24bit with 2^6ms resolution. The change in the
+     * unit facilitates the arrival time computations, as the deltas have 250µs
+     * resolution.
+     */
     public static long getReferenceTime(ByteArrayBuffer fciBuffer)
     {
         byte[] buf = fciBuffer.getBuffer();
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
@@ -55,17 +55,18 @@ import java.util.*;
  * }</pre>
  *
  * @author Boris Grozev
+ * @author George Politis
  */
 public class RTCPTCCPacket
     extends RTCPFBPacket
 {
     /**
      * Gets a boolean indicating whether or not the RTCP packet specified in the
-     * {@link ByteArrayBuffer} that is passed as an argument is a NACK packet or
+     * {@link ByteArrayBuffer} that is passed as an argument is a TCC packet or
      * not.
      *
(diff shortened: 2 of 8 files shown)


src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -111,11 +96,21 @@
     private RTCPTCCPacket.PacketMap incomingPackets;
 
     /**
+     * The media SSRC of the outgoing RTCP TCC packets.
```
bgrozev: The name says "incoming", but the documentation says "outgoing". Also, what is "media SSRC"?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -330,41 +325,42 @@ public void tccReceived(RTCPTCCPacket tccPacket)
...
-
-        lastTimestampUs = referenceTime;
-
-        for (Map.Entry<Integer, Long> entry
-            : tccPacket.getPackets().entrySet())
-        {
-            PacketDetail packetDetail
-                = sentPacketDetails.remove(entry.getKey());
-
             if (packetDetail != null && videoStream != null)
```
bgrozev: Am I missing something, or is videoStream useless in this method?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -330,41 +325,42 @@ public void tccReceived(RTCPTCCPacket tccPacket)
...
         {
-            currentOffsetMs = System.currentTimeMillis();
-        }
-        else
-        {
-            long delta = referenceTime - lastTimestampUs;
-
-            // Detect and compensate for wrap-arounds in base time.
-            if (Math.abs(delta - kBaseTimestampRangeSizeUs) < Math.abs(delta))
+            long arrivalTimeUs = entry.getValue();
```
bgrozev: I think the name "arrivalTimeUs" is a little confusing.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
gpolitis: No, it's not useless. For the outgoing RTCP TCC packets we need to set the source/media SSRC to be the primary SSRC of the incoming video media stream, otherwise Chrome might ignore the TCC packets.

bgrozev: But this method handles incoming TCC, not outgoing. What am I missing?

gpolitis: You're right.. I was thinking about something else. It looks like this is completely useless, thanks for noticing @bgrozev, I'll test it to make sure it doesn't break anything and remove it.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -320,60 +384,64 @@ public void onReceiveBitrateChanged(Collection<Long> ssrcs, long bitrate)
...
     public void tccReceived(RTCPTCCPacket tccPacket)
     {
-        MediaStream videoStream = null;
-        for (MediaStream stream : mediaStreams)
+        RTCPTCCPacket.PacketMap packetMap = tccPacket.getPackets();
+        long previousArrivalTimeMs = -1;
+        for (Map.Entry<Integer, Long> entry : packetMap.entrySet())
         {
-            if (stream instanceof VideoMediaStream)
+            long arrivalTime250Us = entry.getValue();
```
bbaldino: i'm confused by this name now :) what does this represent?

gpolitis: The time unit that is used for the deltas in the TCC packet is 250us. I wonder if there's a better name that I could use.

bbaldino: gotcha, so if this had a value of 1, that represents 250us...if it was 2, it would be 500us, yeah? if that's the case then the name doesn't seem bad to me.

src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
```diff
@@ -410,7 +410,9 @@ private RawPacket doTransform(RawPacket pkt, boolean send)
             {
                 ByteArrayBuffer baf = it.next();
                 int pt = RTCPHeaderUtils.getPacketType(baf);
-                if (pt == RTCPRRPacket.RR || RTCPREMBPacket.isREMBPacket(baf))
+                if (pt == RTCPRRPacket.RR
+                        || RTCPREMBPacket.isREMBPacket(baf)
+                        || RTCPTCCPacket.isTCCPacket(baf))
```
bgrozev: What do you think about adding a "continue" here, or changing the second "if" to an "else if"? Calling remove() twice is illegal (at least for this iterator), and the code is unclear that it doesn't happen.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -330,41 +394,53 @@ public void tccReceived(RTCPTCCPacket tccPacket)
...
             {
-                delta += kBaseTimestampRangeSizeUs;  // Wrap forwards.
+                continue;
             }
 
-            currentOffsetMs += delta / 1000;
-        }
+            if (remoteReferenceTimeMs == -1)
+            {
+                remoteReferenceTimeMs = RTCPTCCPacket.getReferenceTime(
```
bgrozev: The field is named "remoteReferenceTimeMs", but if I'm reading RTCPTCCPacket.getReferenceTime correctly, it is assigned a timestamp with the 250us format

bgrozev: Oh, I'm wrong, I didn't see the "/4"

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
gpolitis: > gotcha, so if this had a value of 1, that represents 250us...if it was 2, it would be 500us, yeah?

right

src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
gpolitis: You're right, I just patched the code.

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -75,25 +76,50 @@ public static boolean isTCCPacket(ByteArrayBuffer baf)
...
+     * {@link RTCPTCCPacket#RTCPTCCPacket(long, long, PacketMap, byte)}.
+     *
+     * @param baf the buffer which contains the RTCP packet.
+     */
+    public static PacketMap getPackets(ByteArrayBuffer baf)
+    {
+        return getPacketsFci(getFCI(baf));
+    }
+
+    public static long getReferenceTime(ByteArrayBuffer fciBuffer)
```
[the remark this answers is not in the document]
gpolitis: Added the docs, thanks @bgrozev.

PR #384, opened 2017-11-23
Add a layer of abstraction to support multiple fec types

bbaldino:
Add an AbstractFECReceiver layer so that we can have both ulpfec and
flexfec implementations.

### DIFF
--- a/src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
+++ b/src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
@@ -40,11 +40,6 @@ public abstract class AbstractFECReceiver
     private static final Logger logger
         = Logger.getLogger(AbstractFECReceiver.class);
 
-    class Statistics {
-        int numRxFecPackets;
-        int numRecoveredPackets;
-    }
-
     /**
      * Statistics for this fec receiver
      */
@@ -278,4 +273,9 @@ public abstract class AbstractFECReceiver
     }
 
     protected abstract RawPacket[] doReverseTransform(RawPacket[] pkts);
+
+    class Statistics {
+        int numRxFecPackets;
+        int numRecoveredPackets;
+    }
 }
--- a/src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
+++ b/src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
@@ -1,0 +1,280 @@
+/*
+ * Copyright @ 2017 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.impl.neomedia.transform.fec;
+
+import org.jitsi.impl.neomedia.transform.*;
+import org.jitsi.service.configuration.*;
+import org.jitsi.service.libjitsi.*;
+import org.jitsi.service.neomedia.*;
+import org.jitsi.util.*;
+
+import java.util.*;
+
+import static org.jitsi.util.RTPUtils.seqNumComparator;
+
+/**
+ * A {@link PacketTransformer} which handles incoming fec packets.  This class
+ * contains only the generic fec handling logic.
+ *
+ * @author bgrozev
+ * @author bbaldino
+ */
+public abstract class AbstractFECReceiver
+    implements PacketTransformer
+{
+    /**
+     * The <tt>Logger</tt> used by the <tt>AbstractFECReceiver</tt> class and
+     * its instances to print debug information.
+     */
+    private static final Logger logger
+        = Logger.getLogger(AbstractFECReceiver.class);
+
+    class Statistics {
+        int numRxFecPackets;
+        int numRecoveredPackets;
+    }
+
+    /**
+     * Statistics for this fec receiver
+     */
+    protected Statistics statistics = new Statistics();
+
+    /**
+     * The SSRC of the fec stream
+     * NOTE that for ulpfec this might be the same as the associated media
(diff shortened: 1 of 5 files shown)


bbaldino:
@bgrozev let me know what you think of this.  this is to set us up for being able to have a flexfec implementation alongside the current ulpfec one.

bgrozev:
This is exciting. Are you just looking for comments on the general approach, or also for minor comments on e.g. style (that would only be useful if this is close to ready for merging)?

src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
```diff
@@ -0,0 +1,280 @@
...
+
+        int pLen = p.getLength();
+        if (pLen > newMedia.getBuffer().length)
+        {
+            newMedia.setBuffer(new byte[pLen]);
+        }
+
+        System.arraycopy(p.getBuffer(), p.getOffset(), newMedia.getBuffer(),
+            0, pLen);
+        newMedia.setLength(pLen);
```
bgrozev: I suppose that these packets always have offset 0 (and I realize this might be existing code being moved), but I think an explicit setOffset(0) would be good to have.

bgrozev:
I wonder if, now that we have a separate cache for packets (both ingress and egress) we shouldn't reuse it instead of keeping our own. I think the cache is positioned very early in the chain, so it should be re-usable for FEC.

src/org/jitsi/impl/neomedia/transform/fec/FECTransformEngine.java
```diff
@@ -118,16 +140,30 @@ public FECTransformEngine()
...
+        AbstractFECReceiver fpt;
         synchronized (fecReceivers)
         {
             fpt = fecReceivers.get(ssrc);
             if (fpt == null)
             {
-                fpt = new FECReceiver(ssrc, incomingPT);
-                fecReceivers.put(ssrc, fpt);
+                try
+                {
```
bgrozev: I'd rather just have a flag which determines whether to use ulpfec or flexfec. When we get to a situation where there's an abundance of FEC implementations to choose from, then we can implement a generic solution :)

src/org/jitsi/util/RTPUtils.java
```diff
@@ -23,6 +23,40 @@
...
+     * A <tt>Comparator</tt> implementation for RTP sequence numbers.
+     * Compares <tt>a</tt> and <tt>b</tt>, taking into account the wrap at 2^16.
+     *
+     * IMPORTANT: This is a valid <tt>Comparator</tt> implementation only if
+     * used for subsets of [0, 2^16) which don't span more than 2^15 elements.
+     *
+     * E.g. it works for: [0, 2^15-1] and ([50000, 2^16) u [0, 10000])
+     * Doesn't work for: [0, 2^15] and ([0, 2^15-1] u {2^16-1}) and [0, 2^16)
+     */
+    public static final Comparator<? super Integer> seqNumComparator
```
bgrozev: Thanks for moving this to the utils class. But we already have RTPUtils.sequenceNumberComparator and I think they implement the same thing.

bbaldino: ah! even better :)

src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
bbaldino: done

bbaldino:
> This is exciting. Are you just looking for comments on the general approach, or also for minor comments on e.g. style (that would only be useful if this is close to ready for merging)?

i'd like comments on everything, as i'd like to get this in independently before taking advantage of it in the flexfec branch.

bbaldino:
> I wonder if, now that we have a separate cache for packets (both ingress and egress) we shouldn't reuse it instead of keeping our own. I think the cache is positioned very early in the chain, so it should be re-usable for FEC.

This definitely sounds interesting, mind if we keep it for a future improvement?  i think we should make sure to look into this though.

src/org/jitsi/impl/neomedia/transform/fec/FECTransformEngine.java
bbaldino: done

src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
```diff
@@ -0,0 +1,281 @@
...
+    implements PacketTransformer
+{
+    /**
+     * The <tt>Logger</tt> used by the <tt>AbstractFECReceiver</tt> class and
+     * its instances to print debug information.
+     */
+    private static final Logger logger
+        = Logger.getLogger(AbstractFECReceiver.class);
+
+    class Statistics {
```
bgrozev: The convention is to place this at the end of the file.

src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
```diff
@@ -0,0 +1,281 @@
...
+     * The SSRC of the fec stream
+     * NOTE that for ulpfec this might be the same as the associated media
+     * stream, whereas for flexfec it will be different
+     */
+    protected long ssrc;
+
+    /**
+     * Allow disabling of handling of ulpfec packets for testing purposes.
+     */
+    protected boolean handleFec = true;
```
bgrozev: Instance fields go after all static fields/code.

bgrozev: Feel free to ignore, since this comes from my old code which doesn't follow the convention...

src/org/jitsi/impl/neomedia/transform/fec/FECTransformEngine.java
```diff
@@ -19,18 +19,23 @@
 import org.jitsi.service.neomedia.*;
 import org.jitsi.util.*;
 
+import java.lang.reflect.*;
```
bgrozev: Is this still needed?

bgrozev:
Left a couple of minor comments, but LGTM.

src/org/jitsi/impl/neomedia/transform/fec/AbstractFECReceiver.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FECTransformEngine.java
bbaldino: nope, fixed

PR #385, opened 2017-11-29
Encoding secondary ssrc refactoring

bbaldino:

PR #387, opened 2017-12-06
fix: Terminate PLIs and FIRs.

gpolitis:


src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
```diff
@@ -410,20 +410,21 @@ private RawPacket doTransform(RawPacket pkt, boolean send)
             {
                 ByteArrayBuffer baf = it.next();
                 int pt = RTCPHeaderUtils.getPacketType(baf);
-                if (pt == RTCPRRPacket.RR
+                if (pt > -1
```
bbaldino: if this is to handle the chrome ssrc of `1`, i think we should use a constant (like `CHROME_DEFAULT_RTCP_MEDIA_SSRC=1` and check for that specifically

bbaldino: ack, just saw this is on payload type, not ssrc.  is this essentially an `isValid` type check?

gpolitis: getPacketType returns an `int` (so that it has enough space to return an error value) but the RTCPFeedbackMessageEvent.PT_PS is a `byte`. I'll add a comment about that.

bgrozev:
Should we change the constants' types to int to avoid the same mistake in the future?

gpolitis:
I think we should, but as part of a different PR.

src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
bbaldino: i was wondering more about this new check of `pt > -1`

gpolitis: `pt` is the return value of `RTCPHeaderUtils.getPacketType`, which returns -1 in case of an error.

bbaldino:
> > Should we change the constants' types to int to avoid the same mistake in the future?
> I think we should, but as part of a different PR.

why not now? this seems like the time it would be done.  would it be a large change?

src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
bbaldino: so if it's an error, do we want to return the packet as-is? or should we return null?

gpolitis:
Well, flipping types can lead to new bugs, if not done carefully, so I'm more worried about regressions than the extend of the required work. On the other hand, without this fix we're running with _very_ broken FIRs/PLIs: the bridge is not terminating them and the clients fail to decrypt them..

bgrozev:
> Well, flipping types can lead to new bugs, if not done carefully

Let's do it carefully then :)

gpolitis:
Sure. I guess my point was that we want this fix released as soon as possible, whereas the other task could wait. But if you guys think that this fix is not urgent, I'm happy to do the job :-)

bgrozev:
I don't mind merging this now, but it's not a new regression and the effects are somewhat limited (presumably NACK works in most cases), so I don't consider it urgent. But I messed up and merged #385 too early, so if we want to update the bridge we need to either temporarily revert or get jitsi-videobridge#576 finished and merged.

src/org/jitsi/impl/neomedia/RTCPConnectorInputStream.java
```diff
@@ -97,7 +97,7 @@ public static void fireRTCPFeedbackMessageReceived(
          */
         if ((length >= 12) && !listeners.isEmpty())
         {
-            byte pt = buffer[offset + 1];
+            int pt = buffer[offset + 1] & 0xFFFF;
```
bgrozev: This should be & 0xFF

bgrozev:
de3a9e LGTM

src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
```diff
@@ -418,7 +418,7 @@ private RawPacket doTransform(RawPacket pkt, boolean send)
                     continue;
                 }
 
-                if (!send)
+                if (!send && pt > -1)
```
imedwei: If I'm reading this patch correctly, is this the only functional difference in this patch?

The PT field should be 7 bits + 1 bit marker and should be accommodated by the 8-bits of the byte with the caveat that any widening operation will have sign extension if we are not careful?

imedwei: Oh now I read the conversation on this change and realize that PT is overloaded with error states in some code paths.  :\

gpolitis: The issue was that `PT_PS` & `PT_TL` were in bytes but the `pt` was an `int`. When `PT_PS` was converted to `int` it had a value of -50 and the code never went into the if block.

PR #389, opened 2017-12-07
Flexfec-03 implementation

bbaldino:
SQUASH


src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -472,6 +472,14 @@ else if (Constants.ULPFEC.equals(encoding))
...
             }
         }
+        else if (Constants.FLEXFEC_03.equals(encoding))
+        {
+            logger.info("Creating FlexFEC transform engine");
+            FECTransformEngine flexFecTransformEngine =
+                new FECTransformEngine(FECTransformEngine.FecType.FLEXFEC_03,
+                    rtpPayloadType, rtpPayloadType, this);
+            setFecTransformEngine(flexFecTransformEngine);
+        }
```
bgrozev: How does that play with ULPFEC? Do we ever want to enable both ulpfec and flexfec (I'm guessing not)?

src/org/jitsi/impl/neomedia/transform/fec/FECTransformEngine.java
```diff
@@ -144,25 +144,32 @@ public FECTransformEngine()
 
         // Assumption: all packets in pkts have the same SSRC
         Long ssrc = findSSRC(pkts);
-        if (ssrc == null)
+        long primarySsrc = getPrimarySsrc(ssrc);
```
bgrozev: If ssrc is null, wouldn't we hit an NPE trying to unbox it?

src/org/jitsi/impl/neomedia/transform/fec/FECTransformEngine.java
```diff
@@ -144,25 +144,32 @@ public FECTransformEngine()
...
+        if (primarySsrc == -1)
         {
             return pkts;
         }
 
         AbstractFECReceiver fpt;
```
bgrozev: Can we rename "fpt" to something more descriptive?

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
bbaldino: yeah my thought was it'd only be one or the other on a stream

src/org/jitsi/impl/neomedia/transform/fec/FECTransformEngine.java
bbaldino: good catch, fixed

src/org/jitsi/impl/neomedia/transform/fec/FECTransformEngine.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03HeaderReader.java
```diff
@@ -0,0 +1,122 @@
...
+            return false;
+        }
+        int maskType = (buffer[flexFecOffset] & 0x40) >> 6;
+        if (maskType != 0)
+        {
+            // We only support flexible (f == 0) mask type
+            return false;
+        }
+
+        int ssrcCount = buffer[flexFecOffset + 8];
```
bgrozev: The byte might have a negative value, perhaps add &0xff?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03HeaderReader.java
```diff
@@ -0,0 +1,122 @@
...
+
+        FlexFec03Mask mask =
+            new FlexFec03Mask(buffer,
+                flexFecOffset + MASK_START_OFFSET_BYTES, seqNumBase);
+
+        // HEADER_MIN_SIZES_BYTES already includes the the size of the smallest
+        // possible packet mask, but maskSizeBytes will include that as well,
+        // so subtract the size of the smallest possible packet mask (2)
+        // here when we want to calculate the total.
+        int flexFecHeaderSize = HEADER_MIN_SIZE_BYTES - 2 + mask.lengthBytes();
```
bgrozev: Do we need an extra check for the length here somewhere? We haven't checked that the buffer is long enough to contain a possible long mask.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
```diff
@@ -0,0 +1,239 @@
...
+ */
+public class FlexFec03Mask
+{
+    private static final int MASK_0_K_BIT = 0;
+    private static final int MASK_0_START_BIT = 1;
+    private static final int MASK_0_END_BIT = 15;
+    private static final int MASK_1_K_BIT = 16;
+    private static final int MASK_1_START_BIT = 17;
+    private static final int MASK_1_END_BIT = 47;
+    private static final int MASK_2_K_BIT = 48;
```
bgrozev: I hope this format changes for the final spec.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
```diff
@@ -0,0 +1,239 @@
...
+            if (sizeBytes > MASK_SIZE_MED)
+            {
+                // Shift to make room for the k bit
+                // We subtract one from the mask start constants here since they
+                // include the k bit
+                LeftToRightBitSet.shiftBitsRight(mask, MASK_2_START_BIT - 3, numBits - 1, 1);
+            }
+            LeftToRightBitSet.shiftBitsRight(mask, MASK_1_START_BIT - 2, numBits - 1, 1);
+        }
+        LeftToRightBitSet.shiftBitsRight(mask, MASK_0_START_BIT - 1, numBits - 1, 1);
```
bgrozev: I really hope the format changes for the final spec. :)

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03HeaderReader.java
bbaldino: do you mean `buffer`, which is passed in? we only read from that, and we do verify that it's at least the minimum size (via the separate 'length' argument...this was somewhat of a port from chrome's code where the separate pointer/length is necessary...here using `buffer.length` is probably better).  but either way, we never write to that buffer, only read from it, and we don't write to a buffer in here.

bbaldino: oh, i think i see what you mean...you mean after parsing the actual mask size, verifying the buffer matches it? unfortunately, we have to read most of the mask in order to even determine the size, but you're right that we don't read past the last k bit (for the large mask), so there is a check we could add there.  i'll add it.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03HeaderReader.java
bbaldino: nice catch!  fixed and added a unit test to verify

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03HeaderReader.java
bbaldino: when looking at this i found that the way i'd set up the interaction between the packet and the header was a bit awkward and made testing this hard...i refactored that a bit and it feels better now (and also checks for this case)

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
bbaldino: a lot of nastiness tucked away in this file :)

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Header.java
```diff
@@ -0,0 +1,41 @@
+package org.jitsi.impl.neomedia.transform.fec;
```
bgrozev: License header

bbaldino: always has to be at leat one...

src/org/jitsi/impl/neomedia/transform/fec/LeftToRightBitSet.java
```diff
@@ -0,0 +1,297 @@
...
+     * LeftToRightBitSet
+     * @param bytes the byte buffer to parse
+     * @param offset the offset at which to start parsing
+     * @param length the length (in bytes) of the chunk to parse
+     * @return a LeftToRightBitSet representing the chunk of the given buffer
+     * based on the given offset and length
+     */
+    public static LeftToRightBitSet valueOf(byte[] bytes, int offset, int length)
+    {
+        int numBits = bytes.length * 8;
```
bgrozev: bytes.length or length?

bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/LeftToRightBitSet.java
```diff
@@ -0,0 +1,297 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.impl.neomedia.transform.fec;
+
+import sun.reflect.generics.reflectiveObjects.*;
+
+import java.util.*;
+import java.util.stream.*;
```
bgrozev: unused imports

bbaldino: 👍 fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    private static LeftToRightBitSet createMaskWithKBits(int sizeBytes, int baseSeqNum, List<Integer> protectedSeqNums)
+    {
+        int numBits = sizeBytes * 8;
+        LeftToRightBitSet mask = new LeftToRightBitSet(numBits);
+        // First create a mask without the k bits
+        for (Integer protectedSeqNum : protectedSeqNums)
+        {
+            int delta = protectedSeqNum - baseSeqNum;
```
bgrozev: wrapping

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
```diff
@@ -0,0 +1,258 @@
...
+     * @param baseSeqNum the base sequence number to use for the mask
+     * @param sortedProtectedSeqNums the sequence numbers this mask should mark
+     * as protected. NOTE: this list MUST be in sorted order
+     * @return the size, in bytes, of the mask that is needed to convey
+     * the given protected sequence numbers
+     */
+    private static int getMaskSizeInBytes(int baseSeqNum, List<Integer> sortedProtectedSeqNums)
+    {
+        int largestSeqNum = sortedProtectedSeqNums.get(sortedProtectedSeqNums.size() - 1);
+        int largestDelta = largestSeqNum - baseSeqNum;
```
bgrozev: wrapping

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
```diff
@@ -0,0 +1,258 @@
...
+     */
+    public List<Integer> getProtectedSeqNums()
+    {
+        LeftToRightBitSet maskWithoutKBits = getMaskWithoutKBits(maskWithKBits);
+        List<Integer> protectedSeqNums = new ArrayList<>();
+        for (int i = 0; i < this.sizeBytes * 8; ++i)
+        {
+            if (maskWithoutKBits.get(i))
+            {
+                protectedSeqNums.add(baseSeqNum + i);
```
bgrozev: wrapping

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
```diff
@@ -0,0 +1,258 @@
...
+     * Get the list of media packet sequence numbers which are marked
+     * as protected in this mask
+     * @return a list of sequence numbers of the media packets marked as
+     * protected by this mask
+     */
+    public List<Integer> getProtectedSeqNums()
+    {
+        LeftToRightBitSet maskWithoutKBits = getMaskWithoutKBits(maskWithKBits);
+        List<Integer> protectedSeqNums = new ArrayList<>();
+        for (int i = 0; i < this.sizeBytes * 8; ++i)
```
bgrozev: Can we loop until sizeBytes*8? The last index (or two, or three) will be invalid due to the k bits. Is maskWithoutKBits.get() just going to return false in this case?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Packet.java
```diff
@@ -0,0 +1,142 @@
...
+        return this.getLength() - this.getHeaderLength() - this.header.size;
+    }
+
+    /**
+     * Get the offset at which the FlexFEC header starts
+     * @return the offset at which the FlexFEC header starts
+     */
+    private int getFlexFecHeaderOffset()
+    {
+        return this.getHeaderLength();
```
bgrozev: I think we need to add the offset here

bgrozev: Or we could use getPayloadOffset()

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Packet.java
```diff
@@ -0,0 +1,142 @@
...
+    public long getProtectedSsrc()
+    {
+        return this.header.protectedSsrc;
+    }
+
+    /**
+     * Returns the size of the FlexFEC payload, in bytes
+     * @return the size of the FlexFEC packet payload, in bytes
+     */
+    public int getPayloadLength()
```
bgrozev: RawPacket already has a method with this signature. Can we add @Override, or rename it to getFlexFecHeaderLength() for example?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,372 @@
...
+     * Inserts packet into an empty slot in pkts, or allocates a new
+     * array and inserts packet into it.  Returns either the original
+     * array (with packet insert) or a new array containing the original contents
+     * of pkts and with packet inserted
+     * @param packet the packet to be inserted
+     * @param pkts the array in which to insert packet
+     * @return the original pkts array with packet inserted, or, a new array
+     * containing all elements in pkts as well as packet
+     */
+    private RawPacket[] insert(RawPacket packet, RawPacket[] pkts)
```
bgrozev: I think that'd be a nice addition to ArrayUtils

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,372 @@
...
+            }
+            if (recoveredPacket.getBuffer().length < RawPacket.FIXED_HEADER_SIZE)
+            {
+                logger.error("Given RawPacket buffer is too small");
+                return false;
+            }
+            // Copy over the recovery RTP header data from the fec packet
+            // (fecPacket contains the RTP header, so we need to copy from it
+            // starting after that)
+            System.arraycopy(fecPacket.getBuffer(), fecPacket.getHeaderLength(),
```
bgrozev: I think this needs to take the fecPacket offset into account

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,372 @@
...
+                recoveredPacket.getBuffer(), 0, RawPacket.FIXED_HEADER_SIZE);
+
+            // Copy over the recovery rtp payload data from the fec packet
+            System.arraycopy(
+                fecPacket.getBuffer(),
+                fecPacket.getHeaderLength() + fecPacket.getFlexFecHeaderSize(),
```
bgrozev: Same

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,372 @@
...
+            if (length > recoveredPacket.getBuffer().length)
+            {
+                logger.error("Length field of recovered packet is larger" +
+                    " than its buffer");
+                return false;
+            }
+            // The length field used in the xor does not include the header
+            // length, but we want to include the fixed header length when
+            // setting the length on the packet object
+            recoveredPacket.setLength(length + RawPacket.FIXED_HEADER_SIZE);
```
bgrozev: The check above doesn't include the fixed header length

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,372 @@
...
+
+            // Set the SN field.
+            RTPUtils.writeShort(recoveredPacket.getBuffer(), 2,
+                (short)missingSequenceNumber);
+
+            // Set the SSRC field.
+            RTPUtils.writeInt(recoveredPacket.getBuffer(), 8,
+                (int)fecPacket.getProtectedSsrc());
```
bgrozev: We have setSequenceNumber and setSSRC helpers

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,372 @@
...
+            RawPacket recoveredPacket = new RawPacket(buf, 0, 1500);
+            if (!startPacketRecovery(this.fecPacket, recoveredPacket))
+            {
+                return null;
+            }
+            for (Integer protectedSeqNum : fecPacket.getProtectedSequenceNumbers())
+            {
+                if (protectedSeqNum != missingSequenceNumber)
+                {
+                    RawPacket mediaPacket = mediaPackets.get(protectedSeqNum);
```
bgrozev: Should we deal with the case of the media packet being cleared from the cache? Or leave it for when we switch to reusing the MediaStream's cache?

src/org/jitsi/util/RTPUtils.java
```diff
@@ -294,4 +294,53 @@ public static boolean isNewerTimestampThan(long a, long b)
     {
         return rtpTimestampDiff(a, b) > 0;
     }
+
+    private static String toHexString(byte b)
+    {
+        char[] hexes
+            = new char[]
```
bgrozev: Can we make these static? Also, missing docs

test/org/jitsi/impl/neomedia/transform/fec/FlexFec03ReceiverTest.java
```diff
@@ -0,0 +1,153 @@
...
+    private void verifyFlexFec(FlexFec03Packet flexFecPacket, Map<Integer, RawPacket> mediaPackets)
+    {
+        List<Integer> protectedSeqNums = flexFecPacket.getProtectedSequenceNumbers();
+        List<RawPacket> protectedMediaPackets = new ArrayList<>();
+        for (Integer protectedSeqNum : protectedSeqNums)
+        {
+            protectedMediaPackets.add(mediaPackets.get(protectedSeqNum));
+        }
+        // We'll treat the first one as the 'lost' packet we'll expect the
+        // FlexFec03Receiver to recover
```
bgrozev: I think it would be nice to test the case of different packets being lost (not just the first one). Should be easy to do with Random with a fixed seed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
bbaldino: it does return false for anything out of range (it only throws if you try and get a negative index).  even still, maybe it would be clearer to specifically use the bit count.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
bbaldino: fixed this to use `maskWithoutKBits.sizeBits()` since we have it

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Packet.java
bbaldino: fixed.  man i hate having to deal with offset.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Packet.java
bbaldino: will rename

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: it looks like `ArrayUtils#add` does something similar, but it doesn't cover the case where there is an existing empty slot that can be used.  not sure if i'd want to mess with `ArrayUtils#add`, should i add this method alongside `ArrayUtils#add`? what do you think?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: cool, fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: hmm, i'm trying to think if this can actually happen.  `AbstractFECReceiver#reverseTransform` gets called, which saves the current fec/media packets.  It then calls `FlexFec03Receiver#doReverseTransform` which looks through the media and fec packets to see if recovery is possible, this is all done synchronously (and `AbstractFECReceiver#reverseTransform` itself is synchronized) so by the time we get here we know that all the media packets we need will be in `mediaPacketss`.

src/org/jitsi/util/RTPUtils.java
bbaldino: fixed

test/org/jitsi/impl/neomedia/transform/fec/FlexFec03ReceiverTest.java
bbaldino: i'm fine with that, as long as the test itself is consistent (i.e. we either always lose the same packet or run through every scenario, each where a different one of the protected packets is lost).

bbaldino: i changed it to test dropping all possible media packets (one by one).  doesn't seem to add a noticeable time difference.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bgrozev: I was thinking of a new method in ArrayUtils, which would use ArrayUtils#add if there is no empty space.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bgrozev: Are we still planning to eventually remove the cache and re-use the existing one, though?

bbaldino: yeah, it seems to me like that would be worth looking into

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: good idea, done.

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
bgrozev: I supposed you've tested this, but this only works if setFecTransformEngine is called before the transform chain is initialized  (if fecTransformEngine is null when the chain is initialized, it will never be added to the chain).

In other places we use a wrapper to solve this problem (TransformEngineWrapper)

src/org/jitsi/util/ArrayUtils.java
```diff
@@ -63,6 +63,25 @@
...
 
+    public static <T> T[] insert(T element, T[] array, Class<T> componentType)
+    {
+        boolean inserted = false;
+        for (int i = 0; i < array.length; ++i)
+        {
+            if (array[i] == null)
+            {
+                array[i] = element;
+                inserted = true;
```
bgrozev: Looks like this inserts the element in ALL null fields of the array. Also, docs.

src/org/jitsi/util/ArrayUtils.java
```diff
@@ -63,6 +63,25 @@
...
+            }
+        }
+
+        if (!inserted)
+        {
+            add(array, componentType, element);
```
bgrozev: We lose track of the new array instance created by add()

src/org/jitsi/util/RTPUtils.java
```diff
@@ -294,4 +294,75 @@ public static boolean isNewerTimestampThan(long a, long b)
     {
         return rtpTimestampDiff(a, b) > 0;
     }
+
```
bgrozev: Are these used anywhere?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+     * The <tt>Logger</tt> used by the <tt>FlexFec03Receiver</tt> class and
+     * its instances to print debug information.
+     */
+    private static final Logger logger
+        = Logger.getLogger(FlexFec03Receiver.class);
+
+    /**
+     * FEC-related statistics
+     */
+    private Statistics statistics;
```
bgrozev: Do we intentionally shadow super.statistics? Looks like super is the one which handles numRxFecPackets

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+
+    public FlexFec03Receiver(long mediaSsrc, byte fecPayloadType)
+    {
+        super(mediaSsrc, fecPayloadType);
+        this.statistics = new Statistics();
+        this.reconstructor = new Reconstructor(mediaPackets);
+    }
+
+    @Override
+    public synchronized RawPacket[] doReverseTransform(RawPacket[] pkts)
```
bgrozev: I think we can keep this protected

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+
+        public void setFecPacket(FlexFec03Packet p)
+        {
+            logger.debug("Have " + mediaPackets.size() + " saved media packets");
+            numMissing = 0;
+            logger.debug("Reconstructor checking if recovery is " +
+                "possible: fec packet " + p.getSequenceNumber() +
+                " protects packets:\n" + p.getProtectedSequenceNumbers());
+            this.fecPacket = p;
+            if (p == null)
```
bgrozev: If p were null it would have already produced a NPE in the debug logs above. Also missing docs on the method.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+         * Initialize the given RawPacket with the RTP header information
+         * and payload from fecPacket
+         * @param fecPacket the FlexFEC packet being used for recovery
+         * @param recoveredPacket the blank RawPacket we're recreating the
+         * recovered packet in
+         * @return true on success, false otherwise
+         */
+        private boolean startPacketRecovery(FlexFec03Packet fecPacket, RawPacket recoveredPacket)
+        {
+            if (fecPacket.getLength() < FlexFec03Packet.FIXED_HEADER_SIZE)
```
bgrozev: Can we reference RawPacket.FIXED_HEADER_SIZE here?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+            {
+                logger.error("Given FlexFEC packet is too small");
+                return false;
+            }
+            if (recoveredPacket.getBuffer().length < RawPacket.FIXED_HEADER_SIZE)
```
bgrozev: 12 bytes is probably not enough. We can require that the recovered packet is at least as long as the FEC packet, which should hold in most cases (unless the FEC packet is invalid).

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+
+        /**
+         * Xor the RTP headers of source and destination
+         * @param source the packet to xor the header from
+         * @param dest the packet to xor the header into
+         */
+        private void xorHeaders(RawPacket source, RawPacket dest)
+        {
+            // XOR the first 2 bytes of the header: V, P, X, CC, M, PT fields.
+            dest.getBuffer()[0] ^= source.getBuffer()[0];
```
bgrozev: The current code creates 'dest' with an offset of 0, but we need to take it into account for 'source'.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+            dest.getBuffer()[1] ^= source.getBuffer()[1];
+
+            // XOR the length recovery field.
+            int length = (source.getLength() & 0xffff) - RawPacket.FIXED_HEADER_SIZE;
```
bgrozev: What purpose does '& 0xffff' serve? It's strange that it is applied before the subtraction.

src/org/jitsi/util/ArrayUtils.java
bbaldino: right you are...fixed

src/org/jitsi/util/ArrayUtils.java
bbaldino: fixed

src/org/jitsi/util/RTPUtils.java
bbaldino: i used them as part of debugging, have found them useful and have had to re-add them a couple times now so thought they would be useful to have.  i can remove/move them if you feel strongly about that.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: nope, think this was probably a copy-paste issue. fixed.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: nice catch :) fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: oops, yeah fixed.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+         * @param sourceOffset the offset in the source packet at which the
+         * payload begins.  Note that this is not necessarily the location
+         * at which the RTP payload starts...for the purpose of FlexFEC,
+         * everything after the fixed RTP header is considered the 'payload'
+         * @param dest the packet to xor the payload into
+         * @param destOffset the offset in the dest packet at which the payload
+         * begins
+         * @param payloadLength the length of the source's payload
+         */
+        private void xorPayloads(RawPacket source, int sourceOffset,
```
bgrozev: If you're passing offsets anyway, might as well pass the byte[] instead of RawPacket

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: changed this to check `< fecPacket.getLength()`, is that what you had in mind?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: fixed (i hate offset)

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: source.getLength() is an int, but the field is only 2 bytes, just being cautious/explicit i suppose.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
```diff
@@ -0,0 +1,341 @@
...
+            }
+            for (Integer protectedSeqNum : fecPacket.getProtectedSequenceNumbers())
+            {
+                if (protectedSeqNum != missingSequenceNumber)
+                {
+                    RawPacket mediaPacket = mediaPackets.get(protectedSeqNum);
+                    xorHeaders(mediaPacket, recoveredPacket);
+                    xorPayloads(
+                        mediaPacket,
+                        RawPacket.FIXED_HEADER_SIZE,
```
bgrozev: mediaPacket.getOffset() + RawPacket.FIXED_HEADER_SIZE

A thought: do you think it's worth artificially adding a non-zero offset to the packets in the tests, to make sure that we didn't miss a place?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: sure, fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: fixed...and yeah i'll take a look at doing that.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Packet.java
```diff
@@ -0,0 +1,133 @@
...
+ * |                                                               |
+ * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+ * |                     ... next in SSRC_i ...                    |
+ * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+ */
+
+public class FlexFec03Packet
+    extends RawPacket
+{
+    protected FlexFec03Header header;
```
bgrozev: Protected? Also docs

bgrozev: And move after statics

bbaldino: will add docs.  what's wrong with protected?

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
```diff
@@ -0,0 +1,262 @@
...
+    private static final int MASK_SIZE_SMALL = 2;
+    private static final int MASK_SIZE_MED = 6;
+    private static final int MASK_SIZE_LARGE = 14;
+
+    public static class MalformedMaskException extends Exception
+    {
+
+    }
+
+    private int sizeBytes;
```
bgrozev: All statics on top

bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+        if (largestDelta <= 14)
+        {
+            return MASK_SIZE_SMALL;
+        }
+        else if (largestDelta <= 45)
+        {
+            return MASK_SIZE_MED;
+        }
+        return MASK_SIZE_LARGE;
```
bgrozev: Do we want to signal an error if the delta is too large?

src/org/jitsi/util/RTPUtils.java
bgrozev: No, I like them. I've used something similar more than once, but never thought to commit it...

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Packet.java
bgrozev: There's no classes that extend FlexFec03Packet, so it could just be private

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Receiver.java
bbaldino: so it turns out a lot of these don't matter, as we copy the media packet into a new RawPacket, so we get an offset of 0...but i do still think it was a good idea, and i've added some tests to verify it now.

src/org/jitsi/impl/neomedia/transform/fec/FlexFec03Mask.java
bbaldino: added an exception here (like we do in the other form of this method) if the delta is too big

bgrozev:
I went through everything and only have a few minor final comments. I leave them here because the inline comments are starting to get confusing:

* Does the initialization of the FECTransformEngine work? See the first thread in the inline comments.
* Add docs for AbstractFECReceiver#doReverseTransform (I find myself having to go check the code to figure out how it's used)
* Removes the re-initialization of 'statistics' in the constructor of FlexFec03Receiver (line 46). You can also make the field final.
* Clean up debug logs in FlexFec03Receiver (and s/Attemping/Attempting/). All logger.debug() should be surrounded by if(logger.isDebugEnabled()) by convention.
* The "Error creating flexfec packet" log in FlexFec03Receiver is confusing
* Add docs for Reconstructor#setFecPacket
* Add licence headers to the test files

Everything else LGTM

bgrozev:
LGTM!

PR #390, opened 2017-12-14
Fixes send-side bandwidth estimations and enables log mining.

gpolitis:
- The first idea is to be able to mine our infra logs for valuable information and be able to correlate different logs together at the conference level or at the user level. We already to that in the JVB code and this PR enables the same logic in LJ.
- Additionally this PR also includes fixes for the send-side bandwidth estimations.

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
@@ -289,13 +289,15 @@ public class RTCPReceiverFeedbackTermination
 
                 if (logger.isTraceEnabled())
                 {
-                    logger.trace("created_report_block," + hashCode()
-                            + "," + System.currentTimeMillis()
-                            + "," + reportBlock.getSSRC()
-                            + "," + reportBlock.getNumLost()
-                            + "," + (reportBlock.getFractionLost() / 256D)
-                            + "," + reportBlock.getJitter()
-                            + "," + reportBlock.getXtndSeqNum());
+                    logger.trace(stream.getDiagnosticContext()
+                            .makeTimeSeriesPoint("created_report_block")
+                            .addKey("rtcp_termination", hashCode())
+                            .addField("ssrc", reportBlock.getSSRC())
+                            .addField("num_lost", reportBlock.getNumLost())
+                            .addField("fraction_lost", reportBlock.getFractionLost() / 256D)
+                            .addField("jitter", reportBlock.getJitter())
+                            .addField("xtnd_seqnum", reportBlock.getXtndSeqNum())
+                            .setTimestampMs(System.currentTimeMillis()));
                 }
             }
         }
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
@@ -294,10 +294,11 @@ public class RTCPReceiverFeedbackTermination
                             .addKey("rtcp_termination", hashCode())
                             .addField("ssrc", reportBlock.getSSRC())
                             .addField("num_lost", reportBlock.getNumLost())
-                            .addField("fraction_lost", reportBlock.getFractionLost() / 256D)
+                            .addField("fraction_lost",
+                                reportBlock.getFractionLost() / 256D)
                             .addField("jitter", reportBlock.getJitter())
-                            .addField("xtnd_seqnum", reportBlock.getXtndSeqNum())
-                            .setTimestampMs(System.currentTimeMillis()));
+                            .addField("xtnd_seqnum",
+                                reportBlock.getXtndSeqNum()));
                 }
             }
         }
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
@@ -487,6 +487,8 @@ public class RTCPTCCPacket
      * missing (not received) packets.
      * @param fbPacketCount the index of this feedback packet, to be used in the
      * "fb pkt count" field.
+     * @param diagnosticContext the {@link DiagnosticContext} to use to print
+     * diagnostic information.
      *
      * Warning: The timestamps for the packets are expected to be in
      * millisecond increments, which is different than the output map produced
@@ -495,9 +497,9 @@ public class RTCPTCCPacket
      * Note: this implementation is not optimized and might not always use
      * the minimal possible number of bytes to describe a given set of packets.
      */
-    public RTCPTCCPacket(long senderSSRC, long sourceSSRC,
-                         PacketMap packets,
-                         byte fbPacketCount)
+    public RTCPTCCPacket(
+            long senderSSRC, long sourceSSRC, PacketMap packets,
+            byte fbPacketCount, DiagnosticContext diagnosticContext)
     {
         super(FMT, RTPFB, senderSSRC, sourceSSRC);
 
@@ -576,6 +578,15 @@ public class RTCPTCCPacket
                     // The small delta is an 8-bit unsigned with a resolution of
                     // 250µs. Our deltas are all in milliseconds (hence << 2).
                     deltas[deltaOff++] = (byte) ((tsDelta << 2) & 0xff);
+                    if (logger.isTraceEnabled())
+                    {
+                        logger.trace(diagnosticContext
+                                .makeTimeSeriesPoint("small_delta")
+                                .addField("seq", seq)
+                                .addField("arrival_time_ms", ts)
+                                .addField("ref_time_ms", nextReferenceTime)
+                                .addField("delta", tsDelta));
+                    }
                 }
                 else if (tsDelta < 8191 && tsDelta > -8192)
                 {
@@ -586,6 +597,15 @@ public class RTCPTCCPacket
                     short d = (short) (tsDelta << 2);
                     deltas[deltaOff++] = (byte) ((d >> 8) & 0xff);
                     deltas[deltaOff++] = (byte) ((d) & 0xff);
+                    if (logger.isTraceEnabled())
+                    {
+                        logger.trace(diagnosticContext
+                                .makeTimeSeriesPoint("large_delta")
+                                .addField("seq", seq)
+                                .addField("arrival_time_ms", ts)
+                                .addField("ref_time_ms", nextReferenceTime)
+                                .addField("delta", tsDelta));
+                    }
                 }
                 else
                 {
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
@@ -510,7 +510,9 @@ public class RTCPTCCPacket
         // Temporary buffer to store the fixed fields (8 bytes) and the list of
         // packet status chunks (see the format above). The buffer may be longer
         // than needed. We pack 7 packets in a chunk, and a chunk is 2 bytes.
-        byte[] buf = new byte[(packetCount / 7 + 1) * 2 + 8];
+        byte[] buf = packetCount % 7 == 0
+            ? new byte[(packetCount / 7) * 2 + 8]
+            : new byte[(packetCount / 7 + 1) * 2 + 8];
         // Temporary buffer to store the list of deltas (see the format above).
         // We allocated for the worst case (2 bytes per packet), which may
         // be longer than needed.
@@ -630,7 +632,7 @@ public class RTCPTCCPacket
         }
 
         off++;
-        if (packetCount % 7 <= 3)
+        if (packetCount % 7 > 0 && packetCount % 7 <= 3)
         {
             // the last chunk was not complete
             buf[off++] = 0;
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
@@ -98,7 +98,7 @@ public class RTCPTCCPacket
      * unit facilitates the arrival time computations, as the deltas have 250µs
      * resolution.
      */
-    public static long getReferenceTime(ByteArrayBuffer fciBuffer)
+    public static long getReferenceTime250us(ByteArrayBuffer fciBuffer)
     {
         byte[] buf = fciBuffer.getBuffer();
         int off = fciBuffer.getOffset();
@@ -144,7 +144,7 @@ public class RTCPTCCPacket
         int baseSeq = RTPUtils.readUint16AsInt(buf, off);
         int packetStatusCount = RTPUtils.readUint16AsInt(buf, off + 2);
 
-        long referenceTime = getReferenceTime(fciBuffer);
+        long referenceTime = getReferenceTime250us(fciBuffer);
 
         // The offset at which the packet status chunk list starts.
         int pscOff = off + 8;
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
@@ -530,9 +530,8 @@ public class RTCPTCCPacket
         off += RTPUtils.writeShort(buf, off, (short) packetCount);
 
         // Set the 'reference time' field
-        off +=
-            RTPUtils.writeUint24(buf, off,
-                                 (int) ((referenceTime >> 6) & 0xffffff));
+        off += RTPUtils.writeUint24(
+                buf, off, (int) ((referenceTime >> 6) & 0xffffff));
 
         // Set the 'fb pkt count' field. TODO increment
         buf[off++] = fbPacketCount;
@@ -576,7 +575,7 @@ public class RTCPTCCPacket
 
                     // The small delta is an 8-bit unsigned with a resolution of
                     // 250µs. Our deltas are all in milliseconds (hence << 2).
-                    deltas[deltaOff++] = (byte ) ((tsDelta << 2) & 0xff);
+                    deltas[deltaOff++] = (byte) ((tsDelta << 2) & 0xff);
                 }
                 else if (tsDelta < 8191 && tsDelta > -8192)
                 {
@@ -595,7 +594,8 @@ public class RTCPTCCPacket
                     // send feedback with such deltas, we should split it up
                     // into multiple RTCP packets. We can't do that here in the
                     // constructor.
-                    throw new IllegalArgumentException("Delta too big, needs new reference.");
+                    throw new IllegalArgumentException(
+                            "Delta too big, needs new reference.");
                 }
 
                 // If the packet was received, the next delta will be relative
@@ -638,12 +638,10 @@ public class RTCPTCCPacket
             buf[off++] = 0;
         }
 
-
         fci = new byte[off + deltaOff];
         System.arraycopy(buf, 0, fci, 0, off);
         System.arraycopy(deltas, 0, fci, off, deltaOff);
     }
-
 
     /**
      * @return the map of packets represented by this {@link RTCPTCCPacket}.
--- a/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
+++ b/src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
@@ -607,27 +607,27 @@ public class RTCPTCCPacket
             // symbol (we've already set 'off' to point to the correct byte).
             //  0 1 2 3 4 5 6 7          8 9 0 1 2 3 4 5
             //  S T <0> <1> <2>          <3> <4> <5> <6>
-            int symbolOffset;
+            int symbolShift;
             switch (seqDelta % 7)
             {
             case 0:
             case 4:
-                symbolOffset = 4;
+                symbolShift = 4;
                 break;
             case 1:
             case 5:
-                symbolOffset = 2;
+                symbolShift = 2;
                 break;
             case 2:
             case 6:
-                symbolOffset = 0;
+                symbolShift = 0;
                 break;
             case 3:
             default:
-                symbolOffset = 6;
+                symbolShift = 6;
             }
 
-            symbol <<= symbolOffset;
+            symbol <<= symbolShift;
             buf[off] |= symbol;
         }
 
--- a/src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
+++ b/src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
@@ -112,6 +112,12 @@ public class TransportCCEngine
     private final Object sentPacketsSyncRoot = new Object();
 
     /**
+     * The {@link DiagnosticContext} to be used by this instance when printing
+     * diagnostic information.
+     */
+    private final DiagnosticContext diagnosticContext;
+
+    /**
      * The time (in milliseconds since the epoch) at which the first received
      * packet in {@link #incomingPackets} was received (or -1 if the map is
      * empty).
@@ -145,8 +151,21 @@ public class TransportCCEngine
     /**
      * Used for estimating the bitrate from RTCP TCC feedback packets
      */
-    private RemoteBitrateEstimatorAbsSendTime bitrateEstimatorAbsSendTime
-        = new RemoteBitrateEstimatorAbsSendTime(this);
+    private final RemoteBitrateEstimatorAbsSendTime bitrateEstimatorAbsSendTime;
+
+    /**
+     * Ctor.
+     *
+     * @param diagnosicContext the {@link DiagnosticContext} of this instance.
+     */
+    public TransportCCEngine(DiagnosticContext diagnosticContext)
+    {
+        Objects.requireNonNull(diagnosticContext);
+        this.diagnosticContext = diagnosticContext;
+        bitrateEstimatorAbsSendTime
+            = new RemoteBitrateEstimatorAbsSendTime(
+                    this, diagnosticContext);
+    }
 
     /**
      * Notifies this instance that a data packet with a specific transport-wide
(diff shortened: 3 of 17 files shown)


bbaldino:
hey george, this looks interesting.  is there a way we could get this logging prefix built into the logger itself?  i.e. some way we could build it into the logger itself (by subclassing it perhaps) so we just had to call `logger.info(...)` and we'd get the prefix for free, rather than having to remember to add `getLoggingId` to every log call?

gpolitis:
That's a good suggestion and I agree with you that a more structured logger would be helpful here, but I didn't intend to add the prefix to every call and there's also a special format that we're trying to enforce here:

    [time series name][keys] [fields] [timestamp]

The logging id is part of the keys and the object printing the logs is free to add more keys, like its own hash code, for example. In the PR, you can see that we prepend the time series name in the logs and right after comes the logging id part.

So the logger shouldn't prepend the logging id at the beginning, it would have to insert it right after the time series name somehow.

bbaldino:
those things (not applying to every call, being inserted in a certain place) still seem achievable without having to manually add it to every call where we want it though i think, no?  we could have `logger.logMetric` or whatever we want to call it maybe?

gpolitis:
I like the idea of a `logMetric` method. How do you imagine its signature?

bbaldino:
> I like the idea of a logMetric method. How do you imagine its signature?

i think it depends on the nature of the metric data a bit, are parts of it static and other parts dynamic?  the static parts we could set up front (e.g. `logger.setMetricInfo(...)`) and the dynamic parts we could make part of the signature.  if there are no static parts, we'd just put it all in the signature e.g. `logger.metric(time series name, key, field, ...)`.  and i think this would still be an improvement over manually concatenating the `getLoggerId` call since it'd be a required part of the `logger.metric` signature.

but do you have some examples of the type of log calls?  i'm not too familiar with what we want it to do, so defining the api is a bit hard.

ibauersachs:
Before you roll something on your own, please investigate if there isn't something builtin in e.g. log4j / log4j2 / logback that we could leverage. NDC/MDC comes to mind, but I'm not sure it's a good fit.

While we're at logging: has there ever been a discussion about using slf4j (if possible along with [Project Lombok](https://projectlombok.org/features/log)'s `@Slf4j` annotation)?

gpolitis:
@ibauersachs correct me if I'm wrong but it looks like the context is per thread (which is fine for servlets because of their threading model) but if you want to access an MDC field from multiple threads (like in JVB/LJ) then you would have to replicate and keep in sync said field in multiple thread contexts.

I like the logging context idea and maybe we should replace the `loggingId` by a `LoggingContext`, but, if the context per thread is a hard requirement in MDC, it's going to be cumbersome to use in our case.

gpolitis:
This new commit is inspired by the NDC/MDC design but, contrary to the NDC/MDC paradigm, there can be different diagnostic context _providers_ in our application. A diagnostic context provider simply provides a _diagnostic context_ in the NDC/MDC sense.

I want the media stream to be one such provider. The diagnostic context is implemented in the `DiagnosticContext` class and it has convenience methods to format time series points. @ibauersachs @bbaldino I'd love to have your feedback in this new approach.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -145,8 +152,22 @@
...
+    /**
+     * Ctor.
+     *
+     * @param diagnosicContextProvider the {@link DiagnosticContextProvider}
+     * that provides the {@link DiagnosticContext} of this instance.
+     */
+    public TransportCCEngine(
+            DiagnosticContextProvider diagnosticContextProvider)
+    {
+        this.diagnosticContextProvider = diagnosticContextProvider;
```
ibauersachs: How about a Objects.requireNonNull(...)?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -417,19 +438,25 @@ public void tccReceived(RTCPTCCPacket tccPacket)
             long arrivalTimeMs = arrivalTime250Us / 4
                 - remoteReferenceTimeMs + localReferenceTimeMs;
 
-            if (logger.isDebugEnabled())
+            DiagnosticContext diagnosticContext
+                = this.diagnosticContextProvider.getDiagnosticContext();
+            if (logger.isDebugEnabled() && diagnosticContext != null)
```
ibauersachs: With the requireNonNull, the null check would be redundant... I think...

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/AimdRateControl.java
```diff
@@ -84,9 +86,10 @@
 
     private float varMaxBitrateKbps;
 
-    public AimdRateControl()
+    public AimdRateControl(DiagnosticContextProvider diagnosticContextProvider)
     {
         reset();
+        this.diagnosticContextProvider = diagnosticContextProvider;
```
ibauersachs: requireNonNull?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/AimdRateControl.java
```diff
@@ -351,6 +354,7 @@ private long multiplicativeRateIncrease(
             long lastMs,
             long currentBitrateBps)
     {
+        //
```
ibauersachs: ?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/AimdRateControl.java
```diff
@@ -456,13 +460,15 @@ public long updateBandwidthEstimate(long nowMs)
                     currentInput.incomingBitRate,
                     nowMs);
 
-        if (isValidEstimate() && logger.isTraceEnabled())
+        DiagnosticContext diagnosticContext = this.diagnosticContextProvider.getDiagnosticContext();
+        if (isValidEstimate()
```
ibauersachs: Not sure how expensive isValidEstimate is, but I'd move isTraceEnabled before it since that is probably more likely to return false

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/InterArrival.java
```diff
@@ -77,12 +79,14 @@ private static long latestTimestamp(long timestamp1, long timestamp2)
...
             double timestampToMsCoeff,
-            boolean enableBurstGrouping)
+            boolean enableBurstGrouping,
+            DiagnosticContextProvider diagnosticContextProvider)
     {
         kTimestampGroupLengthTicks = timestampGroupLengthTicks;
         this.timestampToMsCoeff = timestampToMsCoeff;
         burstGrouping = enableBurstGrouping;
         numConsecutiveReorderedPackets = 0;
+        this.diagnosticContextProvider = diagnosticContextProvider;
```
ibauersachs: requireNonNull

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/InterArrival.java
```diff
@@ -222,11 +226,15 @@ else if (isNewTimestampGroup(arrivalTimeMs, timestamp))
                     = (int)
                         (currentTimestampGroup.size - prevTimestampGroup.size);
 
-                if (logger.isTraceEnabled())
+                DiagnosticContext diagnosticContext = this.diagnosticContextProvider.getDiagnosticContext();
+                if (logger.isTraceEnabled() && diagnosticContext != null)
```
ibauersachs: probably redundant null check

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseDetector.java
```diff
@@ -52,12 +52,15 @@
...
-    public OveruseDetector(OverUseDetectorOptions options)
+    private final DiagnosticContextProvider diagnosticContextProvider;
+
+    public OveruseDetector(OverUseDetectorOptions options, DiagnosticContextProvider diagnosticContextProvider)
     {
         if (options == null)
             throw new NullPointerException("options");
 
         threshold = options.initialThreshold;
+        this.diagnosticContextProvider = diagnosticContextProvider;
```
ibauersachs: requireNonNull

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseDetector.java
```diff
@@ -135,15 +138,18 @@ else if (T < -threshold)
             newHypothesis = true;
         }
 
-        if (newHypothesis && logger.isTraceEnabled())
+        DiagnosticContext diagnosticContext
+            = this.diagnosticContextProvider.getDiagnosticContext();
+        if (diagnosticContext != null && newHypothesis && logger.isTraceEnabled())
```
ibauersachs: probably redunant null check

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
```diff
@@ -110,8 +112,11 @@
 
     private double varNoise;
 
-    public OveruseEstimator(OverUseDetectorOptions options)
+    public OveruseEstimator(
+            OverUseDetectorOptions options,
+            DiagnosticContextProvider diagnosticContextProvider)
     {
+        this.diagnosticContextProvider = diagnosticContextProvider;
```
ibauersachs: requireNonNull

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseEstimator.java
```diff
@@ -250,16 +255,18 @@ public void update(
         prevOffset = offset;
         offset = offset + K[1] * residual;
 
-        if (logger.isTraceEnabled())
+        DiagnosticContext diagnosticContext
+            = this.diagnosticContextProvider.getDiagnosticContext();
+        if (diagnosticContext != null && logger.isTraceEnabled())
```
ibauersachs: redundant null check

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -198,14 +199,21 @@
...
     /**
      * Ctor.
      *
      * @param observer the observer to notify on bitrate estimation changes.
      */
-    public RemoteBitrateEstimatorAbsSendTime(RemoteBitrateObserver observer)
+    public RemoteBitrateEstimatorAbsSendTime(RemoteBitrateObserver observer, DiagnosticContextProvider diagnosticContextProvider)
     {
         this.observer = observer;
+        this.diagnosticContextProvider = diagnosticContextProvider;
```
ibauersachs: requireNonNull

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -435,21 +443,27 @@ public void incomingPacketInfo(
...
         // Shift up send time to use the full 32 bits that inter_arrival
         // works with, so wrapping works properly.
         long timestamp = sendTime24bits << kAbsSendTimeInterArrivalUpshift;
 
         // Convert the expanded AST (32 bits, 6.26 fixed point) to millis.
         long sendTimeMs = (long) (timestamp * kTimestampToMs);
 
+        DiagnosticContext diagnosticContext
+            = diagnosticContextProvider.getDiagnosticContext();
+        if (logger.isTraceEnabled() && diagnosticContext != null)
```
ibauersachs: redundant null check

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorSingleStream.java
```diff
@@ -75,9 +76,15 @@
...
 
-    public RemoteBitrateEstimatorSingleStream(RemoteBitrateObserver observer)
+    private final DiagnosticContextProvider diagnosticContextProvider;
+
+    public RemoteBitrateEstimatorSingleStream(
+            RemoteBitrateObserver observer,
+            DiagnosticContextProvider diagnosticContextProvider)
     {
         this.observer = observer;
+        this.diagnosticContextProvider = diagnosticContextProvider;
```
ibauersachs: requireNonNull

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
```diff
@@ -120,17 +120,27 @@
...
+            RemoteBitrateObserver observer,
+            DiagnosticContextProvider diagnosticContextProvider)
     {
         this.observer = observer;
 
         // Initialize to the default RTP timestamp based RBE.
-        rbe = new RemoteBitrateEstimatorSingleStream(observer);
+        this.rbe = new RemoteBitrateEstimatorSingleStream(
+                observer, diagnosticContextProvider);
+        this.diagnosticContextProvider = diagnosticContextProvider;
```
ibauersachs: move up and requireNonNull

src/org/jitsi/util/DiagnosticContextProvider.java
```diff
@@ -0,0 +1,6 @@
+package org.jitsi.util;
```
ibauersachs: License header

bbaldino:
this definitely looks better, here are my thoughts:
1) i think all the time series names should be using constants
2) why do we need the `DiagnosticsProvider`?  what does sharing a single DiagnosticsContext give us? i thought maybe we initialized it with some value that we want to be shared but it doesn't look like it?
3) it still feels a bit verbose to me...i think a big part of this is having to make the call to get the `DiagnosticsContext` every single time, so getting rid of that would fix it i think?

gpolitis:
@bbaldino do you mean just the time series names? or also the fields and the keys?

bbaldino:
> @bbaldino do you mean just the time series names? or also the fields and the keys?

probably at least the names, but i'd think anything we write (or may write) in more than once place we'd want a constant for, to try and prevent typos (which would result in different keys and defeat the whole purpose of log mining :)

gpolitis:
I now realize that I haven't provided much context here. I'm happy to replace any string literal with constants, but neither the field names nor time series names are supposed to be reused in a Java application, so adding constants for the field names may end up being a bit too verbose without any tangible benefits.

bbaldino:
> I now realize that I haven't provided much context here. I'm happy to replace any string literal with constants, but neither the field names nor time series names are supposed to be reused in a Java application, so adding constants for the field names may end up being a bit too verbose without any tangible benefits.

Ok, if that's the case (and these keys are only used in the one spot) then constants don't seem to important.

src/org/jitsi/impl/neomedia/rtcp/RTCPReceiverFeedbackTermination.java
```diff
@@ -289,13 +289,17 @@ private long getSenderSSRC()
...
                 if (logger.isTraceEnabled())
                 {
-                    logger.trace("created_report_block," + hashCode()
-                            + "," + System.currentTimeMillis()
-                            + "," + reportBlock.getSSRC()
-                            + "," + reportBlock.getNumLost()
-                            + "," + (reportBlock.getFractionLost() / 256D)
-                            + "," + reportBlock.getJitter()
-                            + "," + reportBlock.getXtndSeqNum());
+                    DiagnosticContext diagnosticContext
```
bbaldino: i'm just trying to think if there's a way we can avoid having to get the diagnostics context every time.  it's probably not a big deal, but the extra line each time seems kinda annoying.  if we know it's not going to be null (i assume so since we don't do a null check here?), then what about just changing it to `logger.trace(stream.getDiagnosticsContext().makeTimeSeries...` ?

also, if we only log each key from a single place (like you mentioned about not needing the constants) then does the global 'context' really help us much?  what about each class creating a timeseries object for the timeseries they will log, and then just keeping that around? (rather than calling `makeTimeSeriesPoint` each time?).  So somewhere in this file we'd have:
```
TimeSeries createdReportBlockTimeSeries = diagnosticContext.makeTimeSeries("created_report_block")
```
and then here we'd just have:
```
logger.trace(createdReportBlockTimeSeries.addKey(...).addField(...)
```

just some ideas

gpolitis: I updated the code to take into account your 1st suggestion, thanks @bbaldino!

As for the 2nd suggestion, I believe your motivation here is to factor out as much functionality as possible, but I think what you're proposing would make the code a bit too verbose, as it would require new class fields to keep a reference to the time series objects.

To give a bit more context here, the functionality that this PR introduces is for debugging purposes, so, the way I imagine it, the API should make adding a time series as frictionless as possible and basically it shouldn't be a whole lot different from adding a debug print message. The API needs to be flexible, terse and promote spatial locality of the changes required to add a time series.

To me it looks like that we've achieved these goals, but please, let me know if you think the `TimeSeries` abstraction is necessary here and I'd be happy to take your suggestion.

gpolitis:
@bgrozev @bbaldino may I please summon you for a review? The first commit that adds the `DiagnosticContext` is reviewed already.

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -145,8 +157,21 @@
...
+
+    /**
+     * Ctor.
+     *
+     * @param diagnosicContext the {@link DiagnosticContext} of this instance.
+     */
+    public TransportCCEngine(DiagnosticContext diagnosticContext)
+    {
+        Objects.requireNonNull(diagnosticContext);
+        this.diagnosticContext = diagnosticContext;
```
bgrozev: requireNonNull returns the instance (on success)

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/AimdRateControl.java
```diff
@@ -84,9 +88,11 @@
 
     private float varMaxBitrateKbps;
 
-    public AimdRateControl()
+    public AimdRateControl(DiagnosticContext diagnosticContext)
     {
         reset();
+        Objects.requireNonNull(diagnosticContext);
```
bgrozev: same as above

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/InterArrival.java
```diff
@@ -77,12 +81,15 @@ private static long latestTimestamp(long timestamp1, long timestamp2)
...
             double timestampToMsCoeff,
-            boolean enableBurstGrouping)
+            boolean enableBurstGrouping,
+            DiagnosticContext diagnosticContext)
     {
         kTimestampGroupLengthTicks = timestampGroupLengthTicks;
         this.timestampToMsCoeff = timestampToMsCoeff;
         burstGrouping = enableBurstGrouping;
         numConsecutiveReorderedPackets = 0;
+        Objects.requireNonNull(diagnosticContext);
```
bgrozev: same as above

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/OveruseDetector.java
```diff
@@ -52,12 +54,18 @@
...
+
+    public OveruseDetector(
+            OverUseDetectorOptions options,
+            DiagnosticContext diagnosticContext)
     {
         if (options == null)
             throw new NullPointerException("options");
 
         threshold = options.initialThreshold;
+        Objects.requireNonNull(diagnosticContext);
```
bgrozev: same as above

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
```diff
@@ -205,13 +217,8 @@ public RawPacket reverseTransform(RawPacket pkt)
...
                 usingAbsoluteSendTime = true;
 
-                this.rbe = new RemoteBitrateEstimatorAbsSendTime(observer);
-                if (logger.isTraceEnabled())
-                {
-                    logger.trace("created_rbe," + hashCode()
-                            + "," + System.currentTimeMillis()
-                            + "," + this.rbe.hashCode());
-                }
+                rbe = new RemoteBitrateEstimatorAbsSendTime(
```
bgrozev: formatting

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
```diff
@@ -232,14 +239,9 @@ public RawPacket reverseTransform(RawPacket pkt)
...
-                    rbe = new RemoteBitrateEstimatorSingleStream(observer);
-                    if (logger.isTraceEnabled())
-                    {
-                        logger.trace("created_rbe," + hashCode()
-                                + "," + System.currentTimeMillis()
-                                + "," + this.rbe.hashCode());
-                    }
 
+                    rbe = new RemoteBitrateEstimatorSingleStream(
+                            observer, diagnosticContext);
```
bgrozev: formatting

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
```diff
@@ -266,6 +268,8 @@ public RawPacket reverseTransform(RawPacket pkt)
                 pkt.getPayloadLength(), pkt.getSSRCAsLong());
         }
 
+        // FIXME what if we don't have abs-send-time header extension.
```
bgrozev: There is code just above which waits for 30 such packets and switches back to the SS impl

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
```diff
@@ -120,17 +120,29 @@
      */
     private RemoteBitrateEstimator rbe;
 
+    /**
+     * The {@link DiangosticContext} for this instance.
```
bgrozev: typo in the link

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
```diff
@@ -605,44 +627,42 @@ else if (tsDelta < 8191 && tsDelta > -8192)
             // symbol (we've already set 'off' to point to the correct byte).
             //  0 1 2 3 4 5 6 7          8 9 0 1 2 3 4 5
             //  S T <0> <1> <2>          <3> <4> <5> <6>
-            int symbolOffset;
+            int symbolShift;
```
bgrozev: I remember we discovered a bug in here, but now that I look at it again I can not find it. Was it just a fictitious bug?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -599,4 +560,136 @@ private MediaStream getMediaStream()
...
+        {
+            if (extensionId != -1)
+            {
+                RawPacket.HeaderExtension he
+                    = pkt.getHeaderExtension((byte) extensionId);
+                if (he != null && he.getExtLength() == 2)
+                {
+                    int seq = RTPUtils.readUint16AsInt(
+                            he.getBuffer(), he.getOffset() + 1);
+                    packetReceived(
```
bgrozev: I know the pt is only used in logging, but it's implicitly converted from a byte to an int here

bgrozev:
I left a few minor comments, everything else LGTO. I like the DiagnosticContext approach.

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
gpolitis: How would you like me to reformat the line?

src/org/jitsi/impl/neomedia/rtcp/RTCPTCCPacket.java
gpolitis: Yeah, and our fictitious fix behaved worse than the original version.

gpolitis:
retest this please

gpolitis:
retest this please

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -145,8 +157,20 @@
...
+
+    /**
+     * Ctor.
+     *
+     * @param diagnosicContext the {@link DiagnosticContext} of this instance.
+     */
+    public TransportCCEngine(DiagnosticContext diagnosticContext)
+    {
+        this.diagnosticContext
+            = Objects.requireNonNull(diagnosticContext, "diagnosticContext");
```
bbaldino: any reason not to use the `@NotNull` annotation?  that's what i thought we settled on (@bgrozev ?)

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -604,7 +603,7 @@ public RawPacket transform(RawPacket pkt)
                 if (logger.isDebugEnabled())
                 {
                     logger.debug("rtp_seq=" + pkt.getSequenceNumber()
-                            + ",pt=" + pkt.getPayloadType()
+                            + ",pt=" + RawPacket.getPayloadType(pkt)
```
bgrozev: Ha! Nice fix. I think we should eventually fix the fact that RawPacket#getPayloadType() and RawPacket#getPayloadType(RawPacket) have different return types...

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
```diff
@@ -446,81 +468,19 @@ public void tccReceived(RTCPTCCPacket tccPacket)
     }
 
     /**
-     * Handles RTP packets for this {@link TransportCCEngine}.
+     * Gets the engine which handles outgoing RTP packets for this instance.
      */
-    private class RTPTransformer
-        extends SinglePacketTransformerAdapter
+    public TransformEngine getEgressEngine()
```
bbaldino: just wondering, is there a reason this class can't still implement `TransformEngine` and return these where needed?  is it not possible to work this ingress/egress into the existing transformengine scheme we have everywhere else?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorAbsSendTime.java
```diff
@@ -198,14 +199,25 @@
...
+     * @param diagnosticContext the {@link DiagnosticContext} of this instance.
      */
-    public RemoteBitrateEstimatorAbsSendTime(RemoteBitrateObserver observer)
+    public RemoteBitrateEstimatorAbsSendTime(
+            RemoteBitrateObserver observer,
+            DiagnosticContext diagnosticContext)
     {
         this.observer = observer;
+        this.diagnosticContext
+            = Objects.requireNonNull(diagnosticContext, "diagnosticContext");
```
bbaldino: `@NotNull` ?

src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
gpolitis: I don't recall any decision regarding `@NotNull`, what's the benefit of using that?

gpolitis:
@bbaldino @bgrozev or @ibauersachs, can someone please merge this?

PR #391, opened 2017-12-22
RetransmissionRequesterImpl refactor

bbaldino:
The motive for this refactor was to make this class more unit-testable, which required abstracting out the thread and clock-time related logic.

Some thoughts on this:
1) I'm not really married to the 'delegate' part of the `RetransmissionRequesterDelegate`, but wasn't sure what else to call it, so open to feedback there.
2) I can't decide if it feels odd or not to have both `RetransmissionRequesterImpl` and `RetransmissionRequesterDelegate` implement `RetransmissionRequester`.  This fits with what a delegate would do, but we don't really delegate _all_ calls to the 'delegate' here (the only concrete example being `enable` which is a no-op in `RetransmissionRequesterDelegate`, but it still feels a bit off maybe).
3) I initially tried using java's [ScheduledExecutorService](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html) for the scheduler, and I do think this would be best, but it felt a bit awkward in the unit test.  I do think it would be worth trying to move towards using it in the codebase though.
4) I thought about removing all internal acquisition of the current time from `RetransmissionRequesterDelegate` and passing it in everywhere, but there is a spot that gets the current time from the `reverseTransform` call chain, and I don't think it would be a good idea to change that interface.
5) For the most part, the logic from `RetransmissionRequesterImpl` was copied into `RetransmissionRequesterDelegate`, but I did refactor some things (mostly splitting logic into helper methods).


bgrozev:
On a high level the "delegate" structure seems very weird to me. There's two sorts of functions here: configuring the RetransmissionRequester from the outside (enable() and setSenderSsrc()) and internal methods (the others). These are clearly separate and putting them together in the RetransmissionRequester interface and having two implementations seems overly complicated.

I would suggest to keep the RetransmissionRequester interface as it is, and have Delegate (or whatever name is appropriate) implement whatever it needs (e.g. hasWork(), doWork(), etc). The scheduler never uses the RetransmissionRequester methods anyway, it might as well work directly with the Delegate.

bbaldino:
> On a high level the "delegate" structure seems very weird to me. There's two sorts of functions here: configuring the RetransmissionRequester from the outside (enable() and setSenderSsrc()) and internal methods (the others). These are clearly separate and putting them together in the RetransmissionRequester interface and having two implementations seems overly complicated.
> 
> I would suggest to keep the RetransmissionRequester interface as it is, and have Delegate (or whatever name is appropriate) implement whatever it needs (e.g. hasWork(), doWork(), etc). The scheduler never uses the RetransmissionRequester methods anyway, it might as well work directly with the Delegate.

yeah i agree, and i wasn't really crazy about this either.  changed to use `RecurringRunnable`.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterImpl.java
```diff
@@ -15,108 +15,46 @@
...
     /**
      * Whether this <tt>PacketTransformer</tt> has been closed.
      */
     private boolean closed = false;
 
-    /**
-     * The {@link MediaStream} that this instance belongs to.
-     */
-    private final MediaStream stream;
+    protected final RetransmissionRequesterDelegate retransmissionRequesterDelegate;
```
bgrozev: Docs, formatting, private?

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterImpl.java
```diff
@@ -15,108 +15,46 @@
...
 
-    /**
-     * The SSRC which will be used as Packet Sender SSRC in NACK packets sent
-     * by this {@code RetransmissionRequesterImpl}.
-     */
-    private long senderSsrc = -1;
+    RecurringRunnableExecutor recurringRunnableExecutor = new RecurringRunnableExecutor();
```
bgrozev: Is there anything stopping us from making this static and saving some threads?

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterImpl.java
```diff
@@ -201,369 +79,44 @@ else if (Constants.RTX.equalsIgnoreCase(format.getEncoding()))
...
-
-
-    /**
-     * Implements {@link TransformEngine#getRTPTransformer()}.
-     */
-    @Override
     public PacketTransformer getRTPTransformer()
     {
-        return this;
+        return this.retransmissionRequesterDelegate.getRTPTransformer();
```
bgrozev: I think this hooks up the delagate as the transformer, bypassing RetransmissionRequesterImpl#reverseTransform()

bgrozev: I think it would be best if all TransformEngine and PacketTransformer logic is handled in RetransmissionRequesterImpl, while the delegate can only provide the minimal needed API (e.g. packetReceived(RawPacket))

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java
```diff
@@ -0,0 +1,582 @@
...
+        {
+            if (logger.isTraceEnabled())
+            {
+                Requester nextDueRequester = dueRequesters.get(0);
+                logger.trace(hashCode() + ": Next nack is scheduled for ssrc " +
+                    nextDueRequester.ssrc + " at " +
+                    Math.max(nextDueRequester.nextRequestAt, 0) +
+                    ".  (current time is " + now + ")");
+            }
+            return Math.max(dueRequesters.get(0).nextRequestAt - now, 0);
```
bgrozev: I don't see getDueRequesters doing anything about the order of the requesters it receives, and I think here we want to min across all requesters. I would also add a method that just iterates and returns the minimum as an int to avoid creating a list every time.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java
```diff
@@ -0,0 +1,582 @@
...
+        if (logger.isTraceEnabled())
+        {
+            logger.trace(hashCode() + " running at " + now);
+        }
+        List<Requester> dueRequesters = getDueRequesters(now);
+        if (logger.isTraceEnabled())
+        {
+            logger.trace(hashCode() + " has " + dueRequesters.size() + " due requesters");
+        }
+        List<NACKPacket> nackPackets = createNackPackets(now, dueRequesters);
```
bgrozev: Do we want to hold the lock between getDueRequesters and createNackPackets?

bgrozev: I'm worried about a thread calling Requester.received() while we're creating nack packets

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java
```diff
@@ -0,0 +1,582 @@
...
+     */
+    public static final int RE_REQUEST_AFTER_MILLIS = 150;
+
+    /**
+     * The interval we'll ask the {@link RecurringRunnableExecutor} to check back
+     * in if there is no current work
+     * TODO(brian): i think we should actually be able to get rid of this and
+     * just rely on scheduled work and the 'work ready now' callback
+     */
+    public static final long WAKEUP_INTERVAL_MILLIS = 1000;
```
bgrozev: Just a heads-up: there's currently no way to make the executor sleep indefinitely, and any value returned from getTimeUntilNextRun() that is <=0 triggers immediate execution.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterImpl.java
bbaldino: good idea! yeah i think it seems reasonable that one thread could handle all the nacks, any thoughts on if heavy load might cause a problem there? worst case we could be missing targets for nack transmission.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterImpl.java
bbaldino: > I think this hooks up the delagate as the transformer, bypassing 
> RetransmissionRequesterImpl#reverseTransform()

good catch, fixed

>I think it would be best if all TransformEngine and PacketTransformer logic is handled in > RetransmissionRequesterImpl, while the delegate can only provide the minimal needed API > (e.g. packetReceived(RawPacket))

think i see what you mean, i'll take a shot at this and see how it looks.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java
bbaldino: yep you're right, i was sorting this at one point but i must've lost it amongst some other changes.  will fix that.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java
bbaldino: yeah i think you're right.  and this means we'd need to take the lock at the call to `received` (currently in `reverseTransform`), is that something we regularly do in the packet path?  if not might need to look at another way to do it to avoid the locking.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java
bbaldino: good to know...might be a nice future improvement for the executor.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterImpl.java
bbaldino: fixed

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterImpl.java
bbaldino: i think that tweak looks pretty good

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java
bbaldino: i ended up adding a new method `getNextDueRequester` which is used by `getTimeUntilNextRun` and returns the requester (if any) which is due next. The other usage of `getDueRequesters` is the `run` method which doesn't actually care about the order so i left that method as it was.  there's a little duplication there...could have one method that just returned the sorted list, but this way as now we don't build the list every time we want to find out the next time...if you think it'd be better this way i'm fine with it.

src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterImpl.java
bgrozev: I guess heavy load could be a problem, but I doubt that it will be a bottleneck (and I would imagine it would also be a problem for the multiple threads scenario, too).

PR #392, opened 2017-12-22
ref: Allows for more flexibility in track creation.

gpolitis:
Provides https://github.com/jitsi/jitsi-videobridge/pull/589

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
+++ b/src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
@@ -102,11 +102,13 @@ public class RTPEncodingDesc
      */
     private final double frameRate;
 
+    private boolean baseIsInitialized = false;
+
     /**
      * The root {@link RTPEncodingDesc} of the dependencies DAG. Useful for
      * simulcast handling.
      */
-    private final RTPEncodingDesc base;
+    private RTPEncodingDesc base;
 
     /**
      * The {@link MediaStreamTrackDesc} that this {@link RTPEncodingDesc}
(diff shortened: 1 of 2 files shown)


src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -761,6 +755,20 @@ FrameDesc getLastReceivedFrame()
...
     {
+        if (!baseIsInitialized)
+        {
+            if (ArrayUtils.isNullOrEmpty(dependencyEncodings))
+            {
+                base = this;
+            }
+            else
+            {
+                base = dependencyEncodings[0].getBaseLayer();
```
bgrozev: Just a general concern: signaling a circular dependency will result in an infinite loop here. At some layer between the client and this code (maybe jicofo, maybe jvb) we need to ensure this doesn't happen.

bgrozev: Looks like the bridge ensure this.

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -761,6 +755,20 @@ FrameDesc getLastReceivedFrame()
      */
     public RTPEncodingDesc getBaseLayer()
     {
+        if (!baseIsInitialized)
```
bbaldino: can checking `base` for null replace the need for this extra state variable?

gpolitis:
https://github.com/jitsi/jitsi-videobridge/pull/589 has been closed.

PR #395, opened 2018-01-16
fix: Drops RTCP packets which failed to be parsed.

bgrozev:
We observed an instance in which one browser connected to jitsi-videobridge was continually sending invalid RTCP SR packets (with either 'rc' or 'length' being incorrect). The bridge would forward them (and it shouldn't forward SRs at all). This makes it drop RTCP packets when parsing fails instead.


bgrozev:
Jenkins test this please

bgrozev:
Jenkins test this please

gpolitis:
Jenkins test this please

PR #396, opened 2018-01-19
logs: Adds a category to select BWE time series.

gpolitis:


gpolitis:
@bgrozev @bbaldino could you please take a look?

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/AimdRateControl.java
```diff
@@ -249,10 +249,10 @@ private void changeRegion(RateControlRegion region, long nowMs)
 
         if (logger.isTraceEnabled())
         {
-            logger.trace(diagnosticContext
+            logger.trace(Logger.Category.BWE, diagnosticContext
                     .makeTimeSeriesPoint("aimd_region", nowMs)
                     .addKey("aimd", hashCode())
-                    .addField("value", region));
+                    .addField("value", region).toString());
```
bbaldino: is the `toString` needed?  was it always needed, or only now with passing the extra argument?

gpolitis:
test this please

src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/AimdRateControl.java
gpolitis: Unfortunately, it is. The trace/debug/info signature mandates a `String`. If we flip that to be an `Object` it fails to compile as the method classes with the overload that takes a `Throwable`.

bbaldino: ok, i thought if it had a toString method it would be converted implicitly

bgrozev: Are you sure? The trace() methods would have these signatures if we change String to Object:
Object
Object, Throwable
Category, Object
Category, Object, Throwable

gpolitis: You're both right, thanks for pointing this out.

gpolitis:
This needs some more work, I'll open a new PR when it's ready.

PR #398, opened 2018-01-26
logs: Use a separate logger for logging time series.

gpolitis:
With this PR all time series can be handled separately from the general purpose logging logic:

    timeseries.level = ALL
    timeseries.org.jitsi.impl.neomedia.rtp.remotebitrateestimator.RemoteBitrateEstimatorAbsSendTime.level = OFF
    timeseries.useParentHandlers = false
    timeseries.handlers = java.util.logging.FileHandler

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
+++ b/src/org/jitsi/impl/neomedia/rtp/TransportCCEngine.java
@@ -63,10 +63,11 @@ public class TransportCCEngine
         = Logger.getLogger(TransportCCEngine.class);
 
     /**
-     * The {@link Logger} to be used by this instance to print time series.
+     * The {@link TimeSeriesLogger} to be used by this instance to print time
+     * series.
      */
-    private static final Logger timeSeriesLogger
-        = Logger.getTimeSeriesLogger(TransportCCEngine.class);
+    private static final TimeSeriesLogger timeSeriesLogger
+        = TimeSeriesLogger.getTimeSeriesLogger(TransportCCEngine.class);
 
     /**
      * The engine which handles incoming RTP packets for this instance. It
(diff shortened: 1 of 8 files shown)


gpolitis:
@bgrozev, @bbaldino or @ibauersachs, what I'm trying to accomplish with this PR is to direct all time series output to a separate logging handler, separate from the main/general purpose one. I think this is a sensible way to achieve what I want, could you please take a look?

bgrozev:
I like the approach. One suggestion I have is, given that this separate logger is supposed to only log timeseries, to create a little wrapper (i.e. a TimeseriesLogger class) which accepts only TimeSeriesPoint objects to log.
Another question that I have is: is this the right way to distinguish these events -- timeseries vs everything else? It seems reasonable to me, but perhaps worth thinking about.

gpolitis:
> I like the approach. One suggestion I have is, given that this separate logger is supposed to only log timeseries, to create a little wrapper (i.e. a TimeseriesLogger class) which accepts only TimeSeriesPoint objects to log.

Thanks for chiming in @bgrozev! Do you have mean something like this?

```java
class TimeseriesLogger extends Logger
{
    // ...

    public void warn(@NotNull TimeSeriesPoint point)
    {
        super.warn(point.toString());
    }

    // ... 
}
``` 

It may just be me, but this looks a bit redundant because the logger class already accepts a `TimeSeriesPoint` object and it will convert it to string via its `toString()` method internally , unless you imagine it providing some additional functionality that is not provided by the `Logger` class?

> Another question that I have is: is this the right way to distinguish these events -- timeseries vs everything else? It seems reasonable to me, but perhaps worth thinking about.

I'm open to suggestions, this is why I'm putting this out on the wild. That said, it looks good enough to me and it does what I need, which is an easy way to redirect all the time series output to a specific file, allowing me to turn on/off time series on a class basis.

bgrozev:
> Thanks for chiming in @bgrozev! Do you mean something like this?
> 
> class TimeseriesLogger extends Logger

No, I was thinking without "extends", to assure that only TimeSeriesPoints make it to the timeseries log file

gpolitis:
got it, I'll update the code

gpolitis:
@bgrozev I just updated the code, please let me know if you have any further comments or if it looks good to you.

PR #401, opened 2018-02-01
fix: Do not try to create an RTCP TCC packet for more than 200 packets.

bgrozev:


gpolitis:
I haven't had the chance to review the last two commits yet.

bbaldino:
went over this with boris for a while offline and i think i managed to get enough context.  we talked about a few small tweaks, so will merge after he adds those.

bbaldino:
thanks for the changes, lgtm.  will merge when tests pass.

PR #404, opened 2018-03-01
Send an FIR request when a non-simulcast MST becomes active

dwilson6:
This fixes an issue where a stream sometimes wouldn't show up for recvonly users. It is a race condition between the MST becoming active and the BitrateController in the videobridge being updated which allows the stream to be forwarded on to other endpoints. Sometimes the update code runs before the stream becomes active.

Related to these issues: https://github.com/jitsi/jitsi-videobridge/issues/598, https://github.com/jitsi/jitsi-videobridge/issues/553

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
+++ b/src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
@@ -218,7 +218,14 @@ public class MediaStreamTrackDesc
     {
         if (!simulcast)
         {
-            frameDesc.getRTPEncoding().setActive(true);
+            RTPEncodingDesc encoding = frameDesc.getRTPEncoding();
+            // Request an independent frame (FIR) if the stream is just becoming active.
+            if(!encoding.isActive() && pkt.getPayloadLength(true) > 0) {
+                ((RTPTranslatorImpl) mediaStreamTrackReceiver.getStream()
+                    .getRTPTranslator()).getRtcpFeedbackMessageSender()
+                    .sendFIR((int) rtpEncodings[0].getPrimarySSRC());
+            }
+            encoding.setActive(true);
             return;
         }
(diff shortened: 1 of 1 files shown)


src/org/jitsi/impl/neomedia/rtp/MediaStreamTrackDesc.java
```diff
@@ -218,7 +218,14 @@ void update(
     {
         if (!simulcast)
         {
-            frameDesc.getRTPEncoding().setActive(true);
+            RTPEncodingDesc encoding = frameDesc.getRTPEncoding();
```
gpolitis: While this looks sensible (with a minor modification: this should be done only if the received packet is not the start of a key frame), I'm wondering whether these checks are needed or not. I believe the webrtc team has confirmed that when a suspended stream is resumed, it must restart with a key frame.

dwilson6: I was seeing issues with creating a conference, having a second user join (recvonly), and immediately having the first user adding a stream after ICE is connected. Sometimes VideoChannel.setRtpEncodingParameters (for user 1) would call bitrateController.update (for user 2) which would fail to send an FIR in SimulcastController because the MST was not active yet. This is with a non-simulcast stream. It appears there's a race condition between the setRtpEncodingParameters path and the MST becoming active.

For the minor modification is this what it would be: `if(!encoding.isActive() && pkt.getPayloadLength(true) > 0 && !frameDesc.isIndependent())`?

gpolitis:
We've recently had to change the way we do stream suspension detection and stream switching because of some transport level changes in Chrome. Could you please test again with JVB@1063 and let us know if you still need this?

dwilson6:
We just upgraded our jvb/libjitsi and it appears this change is no longer needed

PR #405, opened 2018-03-01
Minor fixes

bgrozev:


src/org/jitsi/impl/neomedia/AbstractRTPPacketPredicate.java
```diff
@@ -89,10 +89,6 @@ public boolean test(ByteArrayBuffer pkt)
         {
             result = false;
         }
-        if (!result)
-        {
-            logger.debug("Caught a non-RTCP/RTP packet.");
```
bbaldino: is this no longer useful? or was it incorrect/misleading?

bgrozev: "The message would be printed for example every time when an RTP packet
is tested with a predicate which expects RTCP."

PR #410, opened 2018-03-16
Media stream track receiver updates

bgrozev:


bgrozev:
18f40c8 succeeded as part of https://github.com/jitsi/jitsi-videobridge/pull/626 (looking at the logs shows that 18f40c8 was used). I don't know why the tests show up as failed here.

Issue #402, opened 2018-02-14
NPE with new RetransmissionRequesterDelegate Refactor


imedwei:
I'm not entirely sure this isn't my fault but just wanted to give you a heads up that I am seeing NPEs being thrown by the new RetransmissionRequesterDelegate refactor.

```
Feb 13, 2018 11:06:46 PM org.jitsi.util.LoggerImpl log
SEVERE: The invocation of the method org.jitsi.impl.neomedia.transform.RetransmissionRequesterDelegate.run() threw an exception.
java.lang.NullPointerException
	at org.jitsi.impl.neomedia.transform.RetransmissionRequesterDelegate$Requester.notifyNackCreated(RetransmissionRequesterDelegate.java:485)
	at org.jitsi.impl.neomedia.transform.RetransmissionRequesterDelegate.createNackPackets(RetransmissionRequesterDelegate.java:312)
	at org.jitsi.impl.neomedia.transform.RetransmissionRequesterDelegate.run(RetransmissionRequesterDelegate.java:175)
	at org.jitsi.util.concurrent.RecurringRunnableExecutor.run(RecurringRunnableExecutor.java:216)
	at org.jitsi.util.concurrent.RecurringRunnableExecutor.runInThread(RecurringRunnableExecutor.java:292)
	at org.jitsi.util.concurrent.RecurringRunnableExecutor.access$000(RecurringRunnableExecutor.java:36)
	at org.jitsi.util.concurrent.RecurringRunnableExecutor$1.run(RecurringRunnableExecutor.java:328)
```

imedwei:
@bbaldino ^

bgrozev:
I also see some of these on beta. Perhaps because the lock used for 'requests' is released between lines 304 and 312?

bbaldino:
What happens when this occurs?  Does it break everything or is it somewhat harmless?  I'll try and take a look but I don't think it'll be soon.

bgrozev:
Just looking at the code I think it would fail to send a NACK when the condition occurs, but it should immediately run again because the exception prevents nextRequestAt from being updated. So mostly harmless, I think.

PR #411, opened 2018-03-21
Fixes

bgrozev:


bbaldino:
Got a chance to look into this and just wanna think out loud a bit here:

From the stacktrace in #402, we get an NPE on line [485](https://github.com/jitsi/libjitsi/blob/master/src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java#L485) when we pull a request from the `requests` map and try to increment `timesRequested`.  From what I can tell, all access to `requests` is synchronized so doesn't look like we have anything to worry about there, so that leaves `notifyNackCreated` being passed a sequence number that is no longer in the requests map.

Sequence number -> `Requester` entries get removed from the requests map in these places:
* [Requester#received](https://github.com/jitsi/libjitsi/blob/cc0411081d0d8026f24a4dc10b05c60a2fc98ad3/src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java#L392) when an old packet is received.
* [Requester#received](https://github.com/jitsi/libjitsi/blob/cc0411081d0d8026f24a4dc10b05c60a2fc98ad3/src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java#L455) when a big jump is detected and we clear everything
* [Requester#notifyNackCreated](https://github.com/jitsi/libjitsi/blob/cc0411081d0d8026f24a4dc10b05c60a2fc98ad3/src/org/jitsi/impl/neomedia/transform/RetransmissionRequesterDelegate.java#L496) when the max retransmission count is hit.

There are 2 different thread contexts here: the `RecurringRunnable` to retransmit NACKs at some interval and the packet receive context.  `Requester#received` is on the packet receive thread, `notifyNackCreated` is on the `RecurringRunnable` thread.

So it seems like what is happening is: in the middle of some nack retransmission logic, a packet could be received and removed from a `Requester`'s map.

So, after all that...I think that's what you've addressed here, right @bgrozev ?  We sync on the due `Requester` in `createNackPackets` which calls `notifyNackCreated`.  `notifyNackCreated` is already synchronized, but we build the list of missing sequence numbers for that `Requester` above, which allows for it to be modified in between.  Your addition of a synchronized block for that higher up should prevent that from happening.  Does my understanding match yours?

bgrozev:
> Does my understanding match yours?

Yes, I think so. My reasoning was:
All access to `requests` is synchronized, and `notifyNackCreated` is synchronized. Since the request is `null`, something must have removed the entry from `requests` before the call to `notifyNackCreated`. The entry was there when the call to `getMissingSeqNums` was made (line 304), so synchronizing around those two calls (lines 304 and 312) should make that situation impossible. And the fix is safe, because between the two calls there is just some logging an updating a local variable.

PR #422, opened 2018-05-08
Align default port range for default allocation with JVB's default.

guusdk:
For dynamic port allocation of media streams, the videobridge uses a
default port range of 10001-20000 UDP. To avoid confusion, it would be
good if libjitsi uses the same default. Previously, libjitsi defined
5000-6000 UDP as the default range.


guusdk:
This change finds its origin in this mailing list thread: http://lists.jitsi.org/pipermail/dev/2018-May/037818.html

bgrozev:
I haven't verified, but I think that the jitsi desktop client does not override the default, so this would change its behavior.

PR #423, opened 2018-05-23
fix: Chrome 67 simulcast compatibility.

gpolitis:
Provides https://github.com/jitsi/jitsi-videobridge/pull/668

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
+++ b/src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
@@ -28,8 +28,6 @@ import java.util.concurrent.atomic.*;
  * Keeps track of how many channels receive it, its subjective quality index,
  * its last stable bitrate and other useful things for adaptivity/routing.
  *
- * TODO rename to Flow.
- *
  * @author George Politis
  */
 public class RTPEncodingDesc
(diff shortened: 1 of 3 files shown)


gpolitis:
@bgrozev @bbaldino PTAL

src/org/jitsi/impl/neomedia/rtp/RTPEncodingDesc.java
```diff
@@ -63,6 +61,16 @@
...
 
+     /**
+      * The maximum time interval (in millis) an encoding can be considered
+      * active without new frames. This value corresponds to 7.5fps + 65 millis
+      * to compensate for network noise. If the network is clogged and we don't
+      * get a new frame within 200 millis, and if the encoding is being
+      * received, then we will ask for a new key frame (this is done in the
+      * JVB in SimulcastController).
+      */
+    private static final int SUSPENSION_THRESHOLD_MS = 200;
```
bbaldino: this logic will apply to the screenshare stream as well, won't it?  will that framerate work there?

gpolitis: I'll run some tests for that, but this code should not run for non-simulcast streams, and simulcast streams should have at least 7.5fps. If we have a simulcast stream that's lower than 7.5fsp may cause issues

gpolitis: So yeah the code doesn't run in the screen sharing case but I tweaked the timeout in an attempt to avoid problems when Chrome activates simulcast for screen sharing.

gpolitis:
The test failure is unrelated (it's a device disconnect problem on iOS) so this is clear for merging.

PR #426, opened 2018-06-27
feat: Augments the send-side bwe to record some stats.

gpolitis:
This PR introduces some bwe related stats in an effort to understand
how much the loss-based algorithm is impacting our bwe. We record how
much time and how many times we spend in loss-limited, loss-degraded and
loss-free states.


src/org/jitsi/impl/neomedia/rtp/sendsidebandwidthestimation/SendSideBandwidthEstimation.java
```diff
@@ -66,6 +86,31 @@
...
+
+    /**
+     * send_side_bandwidth_estimation.cc
+     */
+    private static final int kDefaultBitrateThresholdKbps = 0;
+
+    /**
+     * Disable the loss experiment by default.
+     */
+    private static final float kDefaultLossExperimentPct = 0;
```
bgrozev: Pct for "percent"? If I'm reading right it's used as a probability (0-1, not 0-100)

src/org/jitsi/impl/neomedia/rtp/sendsidebandwidthestimation/SendSideBandwidthEstimation.java
```diff
@@ -488,4 +566,144 @@ private synchronized void fireBandwidthEstimationChanged(
             second = b;
         }
     }
+
+    public class StatisticsImpl implements Statistics
```
bgrozev: Does this come from webrtc.org?

src/org/jitsi/impl/neomedia/rtp/sendsidebandwidthestimation/SendSideBandwidthEstimation.java
```diff
@@ -488,4 +566,144 @@ private synchronized void fireBandwidthEstimationChanged(
...
+        @Override
+        public void update(long nowMs)
+        {
+            synchronized (SendSideBandwidthEstimation.this)
+            {
+                update(nowMs, null);
+            }
+        }
+
+        // This method is synchronized on SendSideBandwidthEstimation.this.
```
bgrozev: There's no synchronized block. Do you mean that it should always be called while holding the lock? If so, I would make the method synchronized. Or at least make it private.

src/org/jitsi/impl/neomedia/rtp/sendsidebandwidthestimation/SendSideBandwidthEstimation.java
gpolitis: No

src/org/jitsi/impl/neomedia/rtp/sendsidebandwidthestimation/SendSideBandwidthEstimation.java
gpolitis: > Do you mean that it should always be called while holding the lock?

Yes.

> If so, I would make the method synchronized.

👍

src/org/jitsi/impl/neomedia/rtp/sendsidebandwidthestimation/SendSideBandwidthEstimation.java
gpolitis: Probability <> percent and in this particular case we care about the percentage, in some form. What this variable holds is the proportion of lost packets; I felt that kDefaultLossExperimentProportion would be too long and that percentage (abbreviated as Pct) is clear enough. I'll finally rename the variable to kDefaultLossExperimentProportion.

gpolitis:
test this please

src/org/jitsi/impl/neomedia/rtp/sendsidebandwidthestimation/SendSideBandwidthEstimation.java
bgrozev: In this case, can you add documentation, please?

PR #429, opened 2018-07-12
Fixes #362 by reloading native library each time when we try to use video device

### LINKED ISSUE #362: Directshow error 0x80004005
Please read details here: https://github.com/jitsi/jitsi/issues/396

GNUDimarik:
Unfortunatelly I could not to debug native library good enough for making right fix for this issue.
But this hack with reloading library seems to me a good workaround

### DIFF
--- a/src/org/jitsi/impl/neomedia/jmfext/media/protocol/directshow/DSCaptureDevice.java
+++ b/src/org/jitsi/impl/neomedia/jmfext/media/protocol/directshow/DSCaptureDevice.java
@@ -61,7 +61,7 @@ public class DSCaptureDevice
      * This pointer is hold and will be released by <tt>DSManager</tt>
      * singleton.
      */
-    private final long ptr;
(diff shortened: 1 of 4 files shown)


GNUDimarik:
CLA is signed

GNUDimarik:
Any update?

bgrozev:
@ibauersachs any thoughts? It looks OK to me, but I don't understand the code and I can't test.

GNUDimarik:
Testing and steps to reproduce:
You should make a video call between two participatns. Next you should set call on hold and after unhold it. You will not see the video because native library is not reloaded and DirectShow returns error.
With library instance which has the fix video works fine because after reloading native lbrary it recreates DirectShow graph and it works fine,

GNUDimarik:
@bgrozev  @ibauersachs what do you think?

GNUDimarik:
Any update?

PR #432, opened 2018-07-31
fixes plain VP9

Toumassa:
Fixes jitsi/libjitsi#649
Adds key frame detection for VP9 depacketizer and uses it for independent frame detection.

### DIFF
--- a/src/org/jitsi/impl/neomedia/codec/video/vp9/DePacketizer.java
+++ b/src/org/jitsi/impl/neomedia/codec/video/vp9/DePacketizer.java
@@ -15,8 +15,6 @@
  */
 package org.jitsi.impl.neomedia.codec.video.vp9;
 
-import org.jitsi.service.neomedia.*;
-
 /**
  * A depacketizer from VP9.
  * See {@link "https://tools.ietf.org/html/draft-ietf-payload-vp9-02"}
@@ -41,9 +39,24 @@ public class DePacketizer
     private static final byte L_BIT = (byte) (1 << 5);
 
     /**
+     * P bit from the first byte of the Payload Descriptor.
+     */
+    private static final byte P_BIT = (byte) (1 << 6);
+
+    /**
      * I bit from the first byte of the Payload Descriptor.
      */
     private static final byte I_BIT = (byte) (1 << 7);
+    
+    /**
+     * Mask for SID value from Layer Indices byte of the Payload Descriptor.
+     */
+    private static final byte SID_MASK = 0xE;
+    
+    /**
+     * Mask for D value from Layer Indices byte of the Payload Descriptor.
+     */
+    private static final byte D_MASK = 0x1;
 
     /**
      * A class that represents the VP9 Payload Descriptor structure defined
@@ -166,5 +179,54 @@ public class DePacketizer
 
             return (buf[loff] & 0xE) >> 1;
         }
+        
+        /**
+         * Check if the current the packet contains a key frame
+         *
+         * @param buf the byte buffer that holds the VP9 payload.
+         * @param off the offset in the byte buffer where the VP9 payload starts.
+         * @param len the length of the VP9 payload.
+         * @return true if the frame is a key frame, false otherwise
+         */
+        public static boolean isKeyFrame(byte[] buf, int off, int len)
+        {
+            // This packet will have its P bit equal to zero, SID or D
+            // bit (described below) equal to zero, and B bit (described below)
+            // equal to 1
+            
+            if (!isValid(buf, off, len))
+            {
+                return false;
+            }
+
+            // P_BIT must be 0 and B_BIT 1
+            // L_BIT must be 1 to ensure we can do further checks for SID and D
+            if((buf[off] & P_BIT) != 0 || 
+               (buf[off] & B_BIT) == 0 || 
+               (buf[off] & L_BIT) == 0 )
+            {
+                return false;
+            }
+
+            int loff = off + 1;
+            if ((buf[off] & I_BIT) != 0)
+            {
+                loff += 1;
+                if ((buf[off + 1] & (1 << 7)) != 0)
+                {
+                    // extended pid.
+                    loff += 1;
+                }
+            }
+            
+            //SID or D bit equal to zero
+            return 
+              ((buf[loff] & SID_MASK) >> 1) == 0 || (buf[loff] & D_MASK) == 0;
+        }
+    }
+
+    public static boolean isKeyFrame(byte[] buf, int off, int len)
+    {
+        return VP9PayloadDescriptor.isKeyFrame(buf, off, len);
     }
 }
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -4074,6 +4074,7 @@ public class MediaStreamImpl
         }
 
         final byte vp8PT = getDynamicRTPPayloadType(Constants.VP8),
+            vp9PT = getDynamicRTPPayloadType(Constants.VP9),
             h264PT = getDynamicRTPPayloadType(Constants.H264);
 
         if (redBlock.getPayloadType() == vp8PT)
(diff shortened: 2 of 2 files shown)


src/org/jitsi/impl/neomedia/codec/video/vp9/DePacketizer.java
```diff
@@ -166,5 +169,46 @@ public static int getSpatialLayerIndex(byte[] buf, int off, int len)
...
         }
+        /**
+         * Check if the current the packet contains a key frame
+         *
+         * @param buf the byte buffer that holds the VP9 payload.
+         * @param off the offset in the byte buffer where the VP9 payload starts.
+         * @param len the length of the VP9 payload.
+         * @return true if the frame is a key frame, false otherwise
+         */
+        public static boolean isKeyFrame(byte[] buf, int off, int len){
```
gpolitis: Tiny little nit-pick: new-line.

src/org/jitsi/impl/neomedia/codec/video/vp9/DePacketizer.java
```diff
@@ -166,5 +169,49 @@ public static int getSpatialLayerIndex(byte[] buf, int off, int len)
...
+            //bit (described below) equal to zero, and B bit (described below)
+            //equal to 1
+            
+            if (!isValid(buf, off, len))
+            {
+                return false;
+            }
+
+            // P_BIT must be 0 and B_BIT 1
+            if((buf[off] & P_BIT) != 0 || (buf[off] & B_BIT) == 0)
```
gpolitis: I think we need to check for the L_BIT as well.

src/org/jitsi/impl/neomedia/codec/video/vp9/DePacketizer.java
```diff
@@ -166,5 +169,49 @@ public static int getSpatialLayerIndex(byte[] buf, int off, int len)
...
+            {
+                loff += 1;
+                if ((buf[off + 1] & (1 << 7)) != 0)
+                {
+                    // extended pid.
+                    loff += 1;
+                }
+            }
+            //SID or D bit equal to zero
+            return ((buf[loff] & 0xE) >> 1) == 0 || (buf[loff] & 0x1) == 0;
```
gpolitis: Could you please extract the mask constants to something that reads XXX_MASK?

src/org/jitsi/impl/neomedia/codec/video/vp9/DePacketizer.java
Toumassa: As far i understand we should check the L_BIT to ensure that SID and D can be checked below?
Is this correct?
if((buf[off] & P_BIT) != 0 || (buf[off] & B_BIT) == 0 ||  (buf[off] & L_BIT) == 0 )

src/org/jitsi/impl/neomedia/codec/video/vp9/DePacketizer.java
Toumassa: I think this needs to be refactored afterwards, as the same kind of refactor is needed in VP8.depacketizer. 
Having said that, do you mean SID_MASK and D_MASK?

src/org/jitsi/impl/neomedia/codec/video/vp9/DePacketizer.java
gpolitis: Yes 👍

gpolitis: Actually, let me read through this a little bit more carefully.

gpolitis: okay, I read it a bit more carefully, this is the correct condition, thanks!

gpolitis:
Thanks for this work Arthur. Would you be able to test your changes with the modified JVB from this PR https://github.com/jitsi/jitsi-videobridge/pull/735? You can start a VP9 call and use the link-conditioner.sh script to make adaptive last-n kick in.

Toumassa:
> Thanks for this work Arthur. Would you be able to test your changes with the modified JVB from this PR [jitsi/jitsi-videobridge#735](https://github.com/jitsi/jitsi-videobridge/pull/735)? You can start a VP9 call and use the link-conditioner.sh script to make adaptive last-n kick in.

Will test this ASAP. Can you explain quickly how is the link-conditioner.sh is for?

gpolitis:
Actually if you're testing with Jitsi Meet, you can simply change the audio only mode and then switch back to video mode. You can do that multiple times. Every time the video should come back within 1-2 seconds at most. Let me know if you can perform this experiment or if you use a custom client and you need to invoke adaptive last-n.

Toumassa:
Yes, i'll do the test with jitsi-meet

Toumassa:
So here is what i've tested.

1) Applied this PR over the branch gpolitis/rewriting-cleanup
2) Used the the branch gpolitis/rewriting-cleanup of jvb and linked it with this libjitsi
3) forced VP9 with jicofo

=> Testing 2 participants with jitsi-meet no video... ( see sending FIR ... in loop)

4) using vp8 with jicofo
=> Testing 2 participants with jitsi-meetvideo OK

5) Removed the check of L_BIT from libjitsi VP9 depacketizer
6) forced VP9 with jicofo

=> Testing 2 participants with jitsi-meetvideo OK
=> Testing multiple switches between audioonly and video on mode. Takes <1s for the videos to appear
  However i see 10 firs after switching back to video on mode.

So can't understand the logic of testing L_BIT is correct as described in `draft-etf-payload-vp9-06` . Hower it makes the detection of key frame to fail as implemented in this PR.

Any ideas?

gpolitis:
That's exciting, thanks for performing this test Arthur! If the FIR/PLI eventually stop, then it must mean that (eventually) we successfully detect a key frame. How did you observe the FIR/PLIs? Was it from the JVB logs or from the WebRTC internals? If it was from the JVB logs, what did you observe exactly?

gpolitis:
It's great that it's working without checking the L bit, but I'm perplexed and I don't understand why. If the L bit is 0, are we really checking the SID or the D fields? Or are we checking something else and it just happens to work by mere chance. Or can it be that the WebRTC engine forgets to set the L bit? The VP9 payload format is still a draft so there may be some nuances in the code that are not yet described in the draft, so I would look in the code to find answers to these questions. I can't do that right now, but I would like us to understand what's going on here a little bit better before we merge this PR.

gpolitis:
Alternatively, we could adapt the code to use framemarkings. I'm enabling those by default (https://github.com/jitsi/jicofo/pull/335) and will try to adapt the JVB to handle those.

Toumassa:
So precisely this does make the client add specific rtp extension headers to mark a frame as key or whatever. And JVB side we detect key frames based on RTP header rather than VP9 headers. Is this correct?

erotavlas-turbo:
Any news on this? [VP9 support](https://community.jitsi.org/t/vp9-broken/14236/29?u=erotavlas) is extremely important.

bgrozev:
> Any news on this?

libjitsi is no longer used in jitsi-videobridge. It now uses jitsi/jitsi-media-transform

Toumassa:
I possibly rework on this PR when JVB 2.0 (and other components) hits the new stable version, hope soon!

erotavlas-turbo:
It would be great to have jitsi with VP9 as soon as possible during this time of heavy traffic. Any news about the release date for JVB 2.0? Thank you

bbaldino:
> It would be great to have jitsi with VP9 as soon as possible during this time of heavy traffic. Any news about the release date for JVB 2.0? Thank you

JVB 2 can be considered stable at this point (master was moved over months ago, and we've now been running it in prod for a few weeks).  We just haven't changed the stable debian packages over to it yet.

Toumassa:
> JVB 2 can be considered stable at this point (master was moved over months ago, and we've now been running it in prod for a few weeks). We just haven't changed the stable debian packages over to it yet.

cmeng-git:
Committed, but still does not appear in the latest libjitsi master repository?

PR #439, opened 2018-09-04
Fixes

bgrozev:


src/org/jitsi/impl/neomedia/RTPConnectorInputStream.java
```diff
@@ -751,6 +745,17 @@ private void runInReceiveThread()
             {
                 receive(p);
             }
+            catch (SocketTimeoutException ste)
```
gpolitis: I'm confused because in the comment it says that we need to handle these exceptions but we no longer give the calling code an opportunity to handle these exceptions (because of the continue, there is no re-throw).

gpolitis: Thanks for the offline discussion Boris. Continuing actually handles the exception.

PR #447, opened 2018-12-17
Use PacketQueue in OutputDataStreamImpl

mstyura:
**TL; DR**: 
Use [`PacketQueue`](https://github.com/jitsi/ice4j/blob/master/src/main/java/org/ice4j/util/PacketQueue.java) instead of manual queue implementation in [`OutputDataStreamImpl`](https://github.com/jitsi/libjitsi/blob/master/src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java).

**Problem**:
There is a thread per `OutputDataStreamImpl` instance, and there are `2` `OutputDataStreamImpl` instances per `JVB` conference - one for `RTCP` and one for `RTP`. So, number of threads to serve all `OutputDataStreamImpl` instances is linear to the number of conferences.

**Solution**:
There [was a plan](https://github.com/jitsi/ice4j/blob/ce42c77b827a36dd056267aac75fbcbb5e93b556/src/main/java/org/ice4j/util/PacketQueue.java#L28) to replace manual queue within [OutputDataStreamImpl](https://github.com/jitsi/libjitsi/blob/13c82500ac35979df2d83dd48ff4082a61ded7de/src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java#L99) with [`PacketQueue`](https://github.com/jitsi/ice4j/blob/ce42c77b82/src/main/java/org/ice4j/util/PacketQueue.java) from `ice4j`. So this PR just implements that plan. Since than `PacketQueue` implementation were optimized to use thread borrowed from `ExecutorService` and the borrowed thread is not blocked when `PacketQueue` is empty. As a result many [thousands of `PacketsQueue` instances can share single thread](https://github.com/jitsi/ice4j/blob/ce42c77b827a36dd056267aac75fbcbb5e93b556/src/test/java/org/ice4j/util/PacketQueueTests.java#L267).
As a result this PR not only implement the original plan to replace manual queue, but also reduce overall thread usage within `JVB` instance, which results in better performance and lower packet latency.

**Pros**:
1. Simpler implementation using `PacketQueue` instead of manual queue implementation.
2. Number of thread required to run `N` `OutputDataStreamImpl` is reduced from `N` to number of cores. The `N` is doubled number of conferences hosted on `JVB` (`RTP` + `RTCP`). So, for `50` conferences there was created `100` threads to handle `OutputDataStreamImpl` items.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>ice4j</artifactId>
-      <version>2.0.0-20181217.172038-21</version>
+      <version>2.0.0-20190111.035544-25</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>ice4j</artifactId>
-      <version>2.0.0-20190111.035544-25</version>
+      <version>2.0.0-20190115.143045-26</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>ice4j</artifactId>
-      <version>2.0.0-20181213.100259-20</version>
+      <version>2.0.0-20181217.172038-21</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/rtp/translator/OutputDataStreamImpl.java
@@ -68,7 +68,7 @@ class OutputDataStreamImpl
     private static final ExecutorService writeExecutor
         = ExecutorUtils.createForkJoinPool(
             Runtime.getRuntime().availableProcessors(),
-            OutputDataStreamImpl.class.getName() + "-");
+            OutputDataStreamImpl.class.getName());
 
     /**
      * Maximum number of {@link RTPTranslatorBuffer} instances to store in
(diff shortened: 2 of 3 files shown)


mstyura:
Good day, @bbaldino!

Could you please review this change which replaces ad-hoc queue implementation in `OutputDataStreamImpl` with `PacketQueue`. It not only simplifies code inside `OutputDataStreamImpl`, but also reduces number of active threads used.
I know this repo is going to be deprecated soon, but I hope I still can push several patches (this and one more) to it. I believe these changes are very useful for existing users of jvb.

Thank you very much, Brian!

mstyura:
Good day, @bbaldino!

Is there any chance you'll have time to review this PR today?

Thank you very much!

bbaldino:
@mstyura I'm sorry but I'm really busy with some work right now, I don't think I'll get to these more significant PRs soon.

mstyura:
@bbaldino could you please suggest anyone from Jitsi team who is able to review these changes instead of you? Thank you!

mstyura:
Hi, @bgrozev! How are you? Could you please review this PR when you have time?
Thank you very much in advance!

mstyura:
@bgrozev thank you for answering!

> we are actively working on Brian's version of the bridge, which doesn't have this code at all, and we expect to start using it in a few weeks. 

Yes, I'm briefly following Brian's work, but  though that this small PR will still be useful anyway (at least for 2-3 month for me and other users of current bridge). 
I didn't expect new bridge so soon: from my experience when something is rewritten it then takes some unexpectedly long time to stabilize until same stability as existing version.

> Because of this I don't want to spend time on this PR right now. Am I missing something?

Nothing, just let me know if you wouldn't have time next 2 week, I could probably use private build with desired changes then.

> Btw, I often miss gh notifications or see them a few days later, feel free to also send me an email if you want my attention on something.

Sounds good! Maybe I could even send you a message in Hangouts when I'll have something else to review or you prefer using emails?

bgrozev:
I won't be able to spend time on it in the next 2 weeks. It's a small change set, but it is major.

Hangouts is also fine

PR #453, opened 2018-12-27
Micro optimizations on hotpaths

mstyura:
**TL; DR**:
Micro-optimization applied on hotpath's found by profiling.

**Problem**:
1. According to Oracle Visual VM there is small but noticeable overhead introduced by method references used in [`reverseTransform`](https://github.com/jitsi/libjitsi/blob/cc9b594edc504c08f79f6d5a767f255671f03693/src/org/jitsi/impl/neomedia/transform/SinglePacketTransformer.java#L128) and [`transform`](https://github.com/jitsi/libjitsi/blob/cc9b594edc504c08f79f6d5a767f255671f03693/src/org/jitsi/impl/neomedia/transform/SinglePacketTransformer.java#L151);
1. A memory traffic caused by allocating new nodes by `LinkedBlockingQueue` in the [`RTPConnectorInputStream`](https://github.com/jitsi/libjitsi/blob/cc9b594edc504c08f79f6d5a767f255671f03693/src/org/jitsi/impl/neomedia/RTPConnectorInputStream.java#L170) could be avoided by using `ArrayBlockingQueue` which pre-allocates memory.

**Solution**:
1. Cache a `Function` objects calling desired method in the object field;
2. Replace `LinkedBlockingQueue` with `ArrayBlockingQueue`.

### DIFF
--- a/src/org/jitsi/impl/neomedia/RTPConnectorInputStream.java
+++ b/src/org/jitsi/impl/neomedia/RTPConnectorInputStream.java
@@ -167,7 +167,7 @@ public abstract class RTPConnectorInputStream<T>
      * garbage collection.
      */
     private final Queue<RawPacket> rawPacketPool
-            = new LinkedBlockingQueue<>(RTPConnectorOutputStream.POOL_CAPACITY);
+            = new ArrayBlockingQueue<>(RTPConnectorOutputStream.POOL_CAPACITY);
 
     /**
      * The background/daemon <tt>Thread</tt> which invokes
--- a/src/org/jitsi/impl/neomedia/transform/SinglePacketTransformer.java
+++ b/src/org/jitsi/impl/neomedia/transform/SinglePacketTransformer.java
@@ -74,18 +74,14 @@ public abstract class SinglePacketTransformer
      * calling overhead on hotpath.
      */
     private final Function<RawPacket, RawPacket> cachedReverseTransform
-        = (pkt) -> {
-        return this.reverseTransform(pkt);
-    };
+        = pkt -> this.reverseTransform(pkt);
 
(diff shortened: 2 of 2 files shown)


mstyura:
Good day, @bgrozev and @gpolitis!

Could any of you please have a look at this tiny pull request?
PS. Please let me known if you are on holiday or vocation now, so I'll not bother you with further pull requests.

Thank you very much!

bbaldino:
Could you share some details on the overhead of the method reference?

mstyura:
@bbaldino the reason I've decided to replace method reference with cached lambda is due to the following snapshot taken with profiler (on build with [ice4j/#168](https://github.com/jitsi/ice4j/pull/168) applied):

*whole top of hotpath*:


*with filter to `PacketTransformer`*


I was unable to understand what is executed behind that lambda, because profiler did not want to show me via `Find in Call Tree/Show subtree` what is invoked from inside of it. Self CPU time suggested that the computationally expensive code was directly inside `org.jitsi.impl.neomedia.transform.SinglePacketTransformer$$Lambda$204.1945188605.get$Lambda()`. 
Maybe the actual code from references method was inlined by JIT and that's why I've observed big self CPU time here or maybe it was due to overhead of invocation via method reference.

When I've replaced method reference with cached lambda I've observed the following profiling session:
*whole top*


*with filter to `PacketTransformer`*:


So, there were no longer any significant time in auto-generated code with cryptic names. The top CPU related to packet transformers were down from `4.3%` to `2.1%` (absolute time is different due to different duration of profling sessions), so I've decided to create this pull request. It does not harm and it will make profiler reports clearer for anyone who is doing profiling.

PR #463, opened 2019-01-25
Fix Jigasi DTLS connection establishment.

gpolitis:
https://github.com/jitsi/libjitsi/commit/505f2d5b646efb2acd18bb83038d095a088a5a71 broke jigasi DTLS connection establishment. This commit, instead of changing how RawPacket.isInvalid() works, it introduces a RawPacket.isRtpRtcp() method -moved from RTCPUtils and patched so as to properly validate RTP packets with sequence number 0-.

It looks like RawPacket.isInvalid() is used to validate other types of packets, notably DTLS packets, even tho it misleadingly bases its checks on minimum lengths of RTP/RTCP packets.


gpolitis:
test this please

PR #464, opened 2019-01-25
Cleanup calls

damencho:
Clearing up some threads after extensive testing with jigasi calling.

### DIFF
--- a/src/org/jitsi/impl/neomedia/ProcessorUtility.java
+++ b/src/org/jitsi/impl/neomedia/ProcessorUtility.java
@@ -90,26 +90,26 @@ public class ProcessorUtility
      */
     public void controllerUpdate(ControllerEvent ce)
     {
-        // If there was an error during configure or
-        // realize, the processor will be closed
-        if (ce instanceof ControllerClosedEvent)
-        {
-            if (ce instanceof ControllerErrorEvent)
-                logger.warn("ControllerErrorEvent: " + ce);
-            else
-                if (logger.isDebugEnabled())
-                    logger.debug("ControllerClosedEvent: " + ce);
-
-            setFailed(true);
-
-            // All controller events, send a notification
-            // to the waiting thread in waitForState method.
-        }
-
         Object stateLock = getStateLock();
 
         synchronized (stateLock)
         {
+            // If there was an error during configure or
+            // realize, the processor will be closed
+            if (ce instanceof ControllerClosedEvent)
+            {
+                if (ce instanceof ControllerErrorEvent)
+                    logger.warn("ControllerErrorEvent: " + ce);
+                else
+                if (logger.isDebugEnabled())
+                    logger.debug("ControllerClosedEvent: " + ce);
+
+                setFailed(true);
+
+                // All controller events, send a notification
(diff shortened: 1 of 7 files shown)


src/org/jitsi/impl/neomedia/ProcessorUtility.java
```diff
@@ -136,18 +136,26 @@ else if (state == Processor.Realized)
...
-                    stateLock.wait();
+                    // don't wait forever, there is some other
+                    // problem where we wait on an already closed
+                    // processor and we nver leave this wait
+                    stateLock.wait(10000);
+
+                    if (processor.getState() < state)
+                    {
+                        // state not reached we consider failure
+                        setFailed(true);
```
bgrozev: If the Processor passes through an intermediary state before the 10s elapse, we will trigger a failure here. I think we should execute this only if we've waited for a certain period (e.g. 10s)

damencho: @bgrozev I just updated it adding new commit. Tell me if you want to move the change to the commit with the initial change and force push so we can keep both commit with both changes separate ... thanks.

src/org/jitsi/impl/neomedia/ProcessorUtility.java
```diff
@@ -136,18 +144,27 @@ else if (state == Processor.Realized)
...
                 try
                 {
-                    stateLock.wait();
+                    // don't wait forever, there is some other
+                    // problem where we wait on an already closed
+                    // processor and we never leave this wait
+                    long startTime = System.nanoTime();
+                    stateLock.wait(WAIT_TIMEOUT);
+                    if (System.nanoTime() - startTime
+                            > TimeUnit.MILLISECONDS.toNanos(WAIT_TIMEOUT))
```
ibauersachs: Since we're on Java 8 now, why don't you use `Instant` and `Duration` instead of juggling manual date conversions?

src/org/jitsi/impl/neomedia/rtp/translator/RTPConnectorImpl.java
```diff
@@ -201,6 +209,11 @@ public synchronized PushSourceStream getControlInputStream()
     public synchronized OutputDataStreamImpl getControlOutputStream()
         throws IOException
     {
+        if (this.closed)
+        {
+            return null;
```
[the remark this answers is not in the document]
damencho: Well, there were several places in the code where this was already handled that connector can return null and in this case, we skip the writing... it looks safer this way, cause I'm not sure wouldn't IllegalStateException break something else ... 
You are right, but I would insist on merging this and work on the other approach later, because:
1. I want to merge current jitsi, libjitsi and jigasi changes cause we continue seeing losing jigasi instances on meet.jit.si
2. And I continue the work on improving jigasi calling even after these changes, as there are more issues I had found under high load and during that work I can test and work on that.

Thanks for looking at it.

src/org/jitsi/impl/neomedia/ProcessorUtility.java
damencho: I just updated it, does it look ok?

PR #467, opened 2019-02-07
RecorderRtpImpl.java is constructing an instance of the inner class

thedracle:
RTPConnectorImpl, which when constructed creates a new
FECTransformEngine instance, passing in the current value of
mediaStream, which is null and initialized afterwards.

This is probably what is causing the issue in this following forum post
regarding jigasi:
https://community.jitsi.org/t/problems-implementing-rtptanslation-in-jigasi-for-transcriptiongateway/15744

And I am wanting to internally use the old Colibri recording method,
which I realize is no longer supported, but I think this is just plainly
broken, and maybe has other implications.

### DIFF
--- a/src/org/jitsi/impl/neomedia/recording/RecorderRtpImpl.java
+++ b/src/org/jitsi/impl/neomedia/recording/RecorderRtpImpl.java
@@ -328,6 +328,14 @@ public class RecorderRtpImpl
         }
 
         /*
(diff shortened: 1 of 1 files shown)


bgrozev:
Thanks for the contribution. Can you please message me with the info on your CLA? I need to confirm that we have it before I can merge.

PR #470, opened 2019-03-19
ref: Uses jitsi-utils, breaks the ice4j dependency.

bgrozev:


ibauersachs:
Was that supposed to go against the Smack 4.2 branch?

PR #478, opened 2019-06-04
Stores the last received rtp packed and exposes it from audio stream.

damencho:


src/org/jitsi/impl/neomedia/AudioMediaStreamImpl.java
```diff
@@ -176,6 +177,25 @@ public AudioMediaStreamImpl(
...
+     *
+     * @return the time in milliseconds of the last input activity related to
+     * this <tt>AudioMediaStream</tt>
+     * @throws IOException only in case we create input stream and it fails,
+     * as we always pass false to skip creating, should never be thrown.
+     */
+    public long getLastInputActivityTime()
+        throws IOException
+    {
+        RTPConnectorInputStream in = getRTPConnector().getDataInputStream(false);
```
bgrozev: This only takes RTP into account, right? Is this intentional?

damencho: Yes.

PR #482, opened 2019-07-19
Moves ByteArrayBuffer to jitsi-utils

bgrozev:
Makes ByteArrayBuffer an interface in jitsi-utils, and makes the SRTP code use the abstract class (so it can also be reused with a different implementation, and not libjitsi's ByteArrayBuffer).


src/org/jitsi/impl/neomedia/transform/srtp/SRTPCryptoContext.java
```diff
@@ -507,8 +505,8 @@ public void processPacketAESCM(RawPacket pkt)
 
         ivStore[14] = ivStore[15] = 0;
 
-        int payloadOffset = pkt.getHeaderLength();
-        int payloadLength = pkt.getPayloadLength();
+        int payloadOffset = RawPacket.getHeaderLength(pkt.getBuffer(), pkt.getOffset(), pkt.getLength());
+        int payloadLength = pkt.getLength() - payloadOffset;
```
gpolitis: here we could use one of the overloads of `RawPacket.getPayloadLength` if we want to

src/org/jitsi/impl/neomedia/transform/srtp/SRTPCryptoContext.java
```diff
@@ -535,8 +533,8 @@ public void processPacketAESF8(RawPacket pkt)
         ivStore[14] = (byte) (roc >> 8);
         ivStore[15] = (byte) roc;
 
-        int payloadOffset = pkt.getHeaderLength();
-        int payloadLength = pkt.getPayloadLength();
+        int payloadOffset = RawPacket.getHeaderLength(pkt.getBuffer(), pkt.getOffset(), pkt.getLength());
+        int payloadLength = pkt.getLength() - payloadOffset;
```
gpolitis: same as above

src/org/jitsi/impl/neomedia/transform/srtp/SRTPCryptoContext.java
bgrozev: I made this simpler, but I don't want to use `getPayloadLength` because it recalculates the header length (same below).

PR #483, opened 2019-07-23
Debug logging of packet mishandling round 2

gpolitis:
partially reverts 96ffe607. 96ffe607 adds some logging around rtp encodings matching that has some negative side effects. The logging logic is best suited for the caller so it has been/will be moved there.

### DIFF
--- a/src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
+++ b/src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
@@ -800,7 +800,7 @@ public class DePacketizer
                 ", pid=" + getPictureId(buf, off) +
                 ", isExtended=" + hasExtendedPictureId(buf, off, len) +
                 ", hex=" +
-                RTPUtils.toHexString(buf, off, Math.min(len, MAX_LENGTH)) +
+                RTPUtils.toHexString(buf, off, Math.min(len, MAX_LENGTH), false) +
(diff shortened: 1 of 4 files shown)


gpolitis:
> isn't even 1 line per packet going to be a lot?

yes! under certain circumstances we were doing that before (by mistake), and this PR attempts to fix that

bbaldino:
> > isn't even 1 line per packet going to be a lot?
> 
> yes! under certain circumstances we were doing that before (by mistake), and this PR attempts to fix that

Oh, I thought all this did was, in some cases, don't format the logging of the entire packet (but it would still take up at least one line)?

gpolitis:
> > > isn't even 1 line per packet going to be a lot?
> > 
> > 
> > yes! under certain circumstances we were doing that before (by mistake), and this PR attempts to fix that
> 
> Oh, I thought all this did was, in some cases, don't format the logging of the entire packet (but it would still take up at least one line)?

you're right, it does that (i think you're referring to the format parameter that i added to the `toHexString` method, which avoids spaces and newlines for the hex dump), but it also removes the warnings from the mediastreamtrackdesc when they fail to match a packet to an encoding (which is what was filling our disks). now that is done in the adaptivetrackprojection after we've looped over *all* mediastreamtrackdesc instances.

gpolitis:
jenkins, test this please

gpolitis:
jenkins, test this please

PR #487, opened 2019-08-29
Move SRTP core implementation to jitsi-srtp package.

JonathanLennox:
Adapt for API differences.


ibauersachs:
AFAIK the Jitsi Desktop never used the OpenSSL functions. But anyway, I'm also missing changes to the build script that references the new library the necessary OSGi imports.

ibauersachs:
jitsi-srtp seems to be missing OSGi exports, so it won't work with Jitsi Desktop (and maybe Jigasi).

bgrozev:
> jitsi-srtp seems to be missing OSGi exports, so it won't work with Jitsi Desktop (and maybe Jigasi).

Oh yeah, I guess this is what Damian ran into, because I see it fallback to another implementation on mac.

damencho:
So, the previous time I tested jigasi apparently it was not tested with correct jitsi-utils version and everything was failing. 
Now with the correct versions, I tested this PR and it was working fine from a Mac machine.

JonathanLennox:
@ibauersachs: Can you take a look at https://github.com/jitsi/jitsi-srtp/pull/3 and see if would fix the problem for Jitsi Desktop?

JonathanLennox:
And yes, only Linux ever used OpenSSL code previously for srtp.

jitsi-srtp only includes OpenSSL code for 64-bit Linux, and links with OpenSSL 1.1.x, so 32-bit linux and older distros (older than Ubuntu 18.04 or Debian stretch) will now also fall back to the Java crypto implementations.

src/org/jitsi/impl/neomedia/transform/dtls/DtlsPacketTransformer.java
```diff
@@ -702,20 +702,23 @@ private SinglePacketTransformer initializeSRTPTransformer(
                     auth_key_length,
                     RTP_auth_tag_length,
                     cipher_salt_length);
+        org.jitsi.utils.logging2.Logger logger2 = new org.jitsi.utils.logging2.LoggerImpl(getClass().getName());
```
bbaldino: does this same file use both logging2 and logging?

JonathanLennox: Yes.  I didn't want to start pulling the ball of yarn that would be switching all of libjitsi over to logging2.

ibauersachs: Why. WHY? Why aren't we just switching to sl4fj!?
Ditch our own shit! This is all from the time without proper libraries around.

JonathanLennox: This is @bbaldino's new logging framework so I'll tag him in here.

bbaldino: @ibauersachs i don't disagree, but that's a much bigger change than i wanted to get into right now.

ibauersachs: I can't follow. What would be the big change?
In logging(1), the `Category` can be mapped to markers, the levels would be some `sed -i` commands. In logging2 the probably not-so-easy thing is the context, but that could be implemented around slf4j too.

Neustradamus:
Any news on it?

Neustradamus:
@ibauersachs: Thanks!

PR #488, opened 2019-09-04
Switch AST to default to true.

bgrozev:
This has been turned on in the bridge for a long time, and I'd like to remove the code which enables it: https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/java/org/jitsi/videobridge/osgi/JvbBundleConfig.java#L151

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
+++ b/src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
@@ -41,40 +41,6 @@ public class RemoteBitrateEstimatorWrapper
     implements RemoteBitrateEstimator, TransformEngine
 {
     /**
-     * The {@link Logger} used by the {@link RemoteBitrateEstimatorWrapper}
-     * class to print debug information.
-     */
-    private static final Logger logger
-        = Logger.getLogger(RemoteBitrateEstimatorWrapper.class);
-
-    /**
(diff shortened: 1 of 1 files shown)


src/org/jitsi/impl/neomedia/rtp/remotebitrateestimator/RemoteBitrateEstimatorWrapper.java
```diff
@@ -63,7 +63,7 @@
     /**
      * Disable AST RBE by default.
```
JonathanLennox: Comment should be updated to match the definition.

bbaldino:
lgtm, but, is it worth having this parameter at all if it one of the options doesn't work (well)?

bgrozev:
Removing this was satisfying :)

bbaldino:
lgtm but looks like the bridge was checking that param too and will need a change:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/jenkins/jitsi-videobridge/src/main/java/org/jitsi/videobridge/osgi/JvbBundleConfig.java:[151,51] cannot find symbol
  symbol:   variable ENABLE_AST_RBE_PNAME
  location: class org.jitsi.impl.neomedia.rtp.remotebitrateestimator.RemoteBitrateEstimatorWrapper
[INFO] 1 error
```

bgrozev:
This is removed in jitsi/jitsi-videobridge#901

PR #489, opened 2019-09-04
use xml/os utils from jitsi-utils

bbaldino:
depends on https://github.com/jitsi/jitsi-utils/pull/15


bbaldino:
decided not to port this stuff over.

ibauersachs:
Good. Try to get rid of at least OSUtils - most of it should be available via Apache Commons.

PR #494, opened 2019-11-27
build: Make sure OpenSSL is linked in Shared Library.

traud:
Otherwise, the resulting shared library may end-up without a link to libcrypto.so.

### DIFF
--- a/src/native/build.xml
+++ b/src/native/build.xml
@@ -1467,7 +1467,7 @@
       <linkerarg value="-m64" if="cross_64" />
 
       <linkerarg value="-Wl,-z,relro" if="is.running.debian"/>
-      <linkerarg value="-lcrypto"/>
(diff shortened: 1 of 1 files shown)


damencho:
The openssl thing is beeing moved with this https://github.com/jitsi/libjitsi/pull/487, which we didn't have time to fully test with Jitsi Desktop and that's why it is not merged.

traud:
OK, that other Pull Request would need a rebase then.
(a) close this one and wait or
(b) merge this one and rebase the other?

traud:
Because big changes might never come and this Pull Request does not hurt, why not merge this?

bgrozev:
> By the way, why does Jenkins not find my CLA?

It needs to be added, should be good now (for libjitsi)

PR #496, opened 2019-11-28
AMR: Encoder added a Null byte to Frame content.

traud:
Many AMR-WB implementations check the length of the frame, and reject frames with wrong lengths, even if only one frame is in the RTP packet. The solution is to use the right length which is dstLen - 1 = srcLen = the length has not changed.

fixes traud/asterisk-amr#17


bgrozev:
If the correct length is `srcLen` why do we allocate an extra byte in `dst` in the first place?

traud:
Because of that dstI++.

In the destination, Codec Mode Request (CMR), the Another Frame Indicator (F), the Frame Type Index (FT), and the Frame Quality Indicator (Q) get pre-pended. Furthermore, the first byte in the source gets removed. Consequently, everything else is moved 6 bits. However, frame type 8 (of AMR-WB) has six Stuffing Bits. Consequently, the last operation `dst[srcLen] = (byte) (0x03 & s) << 6);` does nothing.

There are two alternatives: (A) Simply cut off that last byte but picky-pack it everywhere (which is OK because we are in Java with its memory management) or (B) change the bit-shifting loop to not do the very last operation on the very last byte. In the latter case, `int dstLen = srcLen;` would be correct. I went for alternative A. Another alternative C would be to rewrite the whole `packetize(.)` and shift in place.

Actually, the current `packetize(.)` is not perfect because it is not agnostic about the frame type. The whole current implementation is hard coded just for frame type 8 although the variable `bitRate` suggests differently. This allows me to leverage a coincidence: `dstLen` stays `srcLen`. Actually, the semantically correct value for the resulting length would be a hard coded value of `61`, because some frame types get bigger (in AMR-WB, just frame type 2), others keep the length. Actually, actually, such an agnostic code would even work for AMR (Narrow-Band), see [this C code…](https://github.com/traud/asterisk-amr/blob/master/codecs/codec_amr.c)

bgrozev:
Sorry for the slow response on this. 

Your fix makes sense, I'm just worried that it makes the code even more confusing than it already is. Can you please add a one line comment to the effect of "we're intentionally keeping the original length and discarding dstLen"?

traud:
> it makes the code even more confusing than it already is

Instead for a comment, I made it correct. Now, `packetize(.)` handles all frame types created by the encoder; it is agnostic about the frame type index. With this, a user is able to change the frame type via `bitRate` as expected. Furthermore, this change allows SID frames created for DTX because of VAD. With these changes, VAD can be enabled via one line of code. I am going to post that in another Pull Request. Successfully tested here with my Sangoma Asterisk module, Google Android (native SIP client), and two Nokia Mobile Phone series.

traud:
Is anything blocking this?

PR #497, opened 2019-11-28
AMR: Send correct timestamp in RTP packets.

traud:
AudioFormat.computeDuration(.) of the Java Media Framework (JMF) returns -1 on default. Therefore, the timestamp in RTP started with 0xffff and decreased by one with each packet. Now, the timestamp increases 320 with every packet (because just a single AMR-WB frame is included in each packet).

fixes jitsi/jitsi#613


bgrozev:
Thanks for the contribution! I left some minor notes, but generally it looks good to me. We'll need the CLA signed before we can merge

traud:
> need the CLA signed before we can merge

Signed it on 11th of November, have it here in front of me. Was it lost somehow? Can I double-check somewhere in a online account that your systems still have it?

PR #499, opened 2019-12-11
H.263+: Remove non-working commented setting.

traud:
The source code contains a line of code which is about the rate-control equation (rc_eq). The given one does not work with the default H.263+ encoder in Debian/Ubuntu and gives the error messages `Unknown function` and `Error parsing`. Because that line of code is commented/disabled anyway, it is simply removed.


damencho:
I think we should delete h263 code in general as it is not used, not sure it is working at all.

PR #500, opened 2019-12-11
H.264: Set the motion-estimation method via private options.

traud:
With FFmpeg 2.7, the motion-estimation method (me_method) was moved from a public symbol to the private options of the underlying encoder interface for libx264. Since FFmpeg 4.0, the public symbol me_method does not exist anymore. Because the latest supported is FFmpeg is 2.8, simply use the private-option interface instead.

### DIFF
--- a/src/org/jitsi/impl/neomedia/codec/video/h264/JNIEncoder.java
+++ b/src/org/jitsi/impl/neomedia/codec/video/h264/JNIEncoder.java
@@ -657,7 +657,6 @@ public class JNIEncoder
                     //"crf" /* constant quality mode, constant ratefactor */, "0",
                     "intra-refresh", intraRefresh ? "1" : "0",
                     "keyint", Integer.toString(keyint),
-                    "motion-est", "hex", /* possible since FFmpeg 2.7 */
                     "partitions", "b8x8,i4x4,p8x8",
                     "preset", preset,
                     "thread_type", "slice",
--- a/src/org/jitsi/impl/neomedia/codec/video/h264/JNIEncoder.java
+++ b/src/org/jitsi/impl/neomedia/codec/video/h264/JNIEncoder.java
@@ -614,7 +614,6 @@ public class JNIEncoder
             FFmpeg.avcodeccontext_add_flags2(avctx,
                 FFmpeg.CODEC_FLAG2_INTRA_REFRESH);
         }
-        FFmpeg.avcodeccontext_set_me_method(avctx, 7);
         FFmpeg.avcodeccontext_set_me_subpel_quality(avctx, 2);
         FFmpeg.avcodeccontext_set_me_range(avctx, 16);
         FFmpeg.avcodeccontext_set_me_cmp(avctx, FFmpeg.FF_CMP_CHROMA);
@@ -658,6 +657,7 @@ public class JNIEncoder
                     //"crf" /* constant quality mode, constant ratefactor */, "0",
                     "intra-refresh", intraRefresh ? "1" : "0",
                     "keyint", Integer.toString(keyint),
+                    "motion-est", "hex", /* possible since FFmpeg 2.7 */
                     "partitions", "b8x8,i4x4,p8x8",
                     "preset", preset,
                     "thread_type", "slice",


traud:
The causing commit in FFmpeg/libav was [4b6b1082](https://code.videolan.org/libav/libav/commit/4b6b1082a73907c7c3de2646c6398bc61320f2c6).

By the way, the encoder OpenH264 does not offer that setting at all. At least I could not find it. This is libx264 specific stuff. Furthermore, I do not know why LibJitsi fiddles around with motion-estimation method at all.

src/org/jitsi/impl/neomedia/codec/video/h264/JNIEncoder.java
```diff
@@ -658,6 +657,7 @@ else if ("h264.profile".equals(k))
                     //"crf" /* constant quality mode, constant ratefactor */, "0",
                     "intra-refresh", intraRefresh ? "1" : "0",
                     "keyint", Integer.toString(keyint),
+                    "motion-est", "hex", /* possible since FFmpeg 2.7 */
```
damencho: Will it cause problems(crashes, errors?) when passing that to FFmpeg < 2.7?

traud: Have not tested that. It is expected that is simply ignored. Consequently, in FFmpeg 2.6 and lower, the motion-estimation method will not be set.

src/org/jitsi/impl/neomedia/codec/video/h264/JNIEncoder.java
```diff
@@ -614,7 +614,6 @@ else if ("h264.profile".equals(k))
             FFmpeg.avcodeccontext_add_flags2(avctx,
                 FFmpeg.CODEC_FLAG2_INTRA_REFRESH);
         }
-        FFmpeg.avcodeccontext_set_me_method(avctx, 7);
```
damencho: This will need and changing the natives at some point, removing that method...

traud: Well, one could keep them just for older FFmpeg. For FFmpeg 4, I created conditionals in that other Pull Request.

src/org/jitsi/impl/neomedia/codec/video/h264/JNIEncoder.java
traud: Does this block this Pull Request? Shall I test something? This Pull Request blocks another Pull Request of mine. 

By the way, I wonder why the motion-estimation method is changed at all. Are you able to ask the original contributor why that is done? Why not dropping this all together.

damencho: It was Lyubomir who did that in the first place, as I remember and he spent time experimenting with some h264 options, but this was long time ago ... I suppose by that time the files were still in jitsi main repo and that was prior ffmpeg 1.0...  So probably that information is already lost/forgotten if there is nothing mentioned in the commit log (which should be imported from the svn repo).

traud: Well, I just found your commit https://github.com/jitsi/jitsi/commit/d6dd75f4a0790fd4da3d927939d1f1f07ecc137b back in the year 2008 which added all this.

traud: What about simply dropping it?

damencho: > Well, I just found your commit [jitsi/jitsi@d6dd75f](https://github.com/jitsi/jitsi/commit/d6dd75f4a0790fd4da3d927939d1f1f07ecc137b) back in the year 2008 which added all this.

You see, I told you it is lost, even I don't remember it :) I just remember that by that time I was also looking at some posts of people with recommended settings for h264 encoding for realtime video.

> What about simply dropping it?

As you had already investigated it and saw it is not used and you have more experience with the latest ffmpeg. I'm ok with that.

traud: Well, currently it is used. I just tweaks x264 to a state nobody knows whether it provides any benefit. In OpenH264, it is not used and ignored anyway.

damencho: > Well, currently it is used. I just tweaks x264 to a state nobody knows whether it provides any benefit. In OpenH264, it is not used and ignored anyway.

If it is not used in openh264, let's drop it.

traud: Wupps. This was not the commit. Here is the one which changed the method to ‘hex’: https://github.com/jitsi/jitsi/commit/21d1855a1463f138113710702bfc7ca43359970a. Anyway, looks like it was changed because there was a not-so-perfect value before.

traud: I dropped it anyway. Now, anyone can blame me.

traud: Nobody to blame: The default = `medium` preset of the encoder library x264 (and the wrapper in FFmpeg) is `X264_ME_HEX` anyway. We can drop it. Done with the commit below.

traud:
Anything blocking this? Just to summarize:
- OpenH264 ignored and ignores that parameter
- x264 is using that value as default, therefore no need to set it

Consequently, that code can simply be removed as done in the latest commit.

PR #502, opened 2019-12-11
H.264: Remove non-working setting rate-control equation.

traud:
The rate-control equation (rc_eq) cannot be set for this encoder, at least not in the default encoder for Debian/Ubuntu. Furthermore, I am not aware of any H.264 encoder which allows this. FFmpeg ignores unknown options silently. Because FFmpeg 4 moved that option to the private options and removed that public symbol completely, and because that option does not do anything in H.264, it is simply removed.


traud:
The causing commit in FFmpeg/libav was [a75c2eb2](https://code.videolan.org/libav/libav/commit/a75c2eb25a62105c09b48521aef429dc8a231637).

Tested with x264. Not fully sure about OpenH264. However, I do not know why LibJitsi fiddles around with rate-control equation at all.

Issue #431, opened 2018-07-16
Can not connect to camera in Windows environment( x86 and x64 / H264, VP8, VP9 all do not work). libjitsi is causing, jitsi is not a problem.


spf-tanabe:
## Description
*  When connecting to the camera with Nightly Builds (x86 and x64), the screen turns black
 * ![image](https://user-images.githubusercontent.com/41299235/42767761-cf9d7a42-8958-11e8-8ebb-eca4c73614a8.png)
* Reproduce with all the following builds
  * https://download.jitsi.org/jitsi/nightly/windows/
```
jitsi-2.11.5604-x86.exe           2018-04-14 03:23   62M
jitsi-2.11-nightly.latest-x86.exe 2018-04-14 03:23   62M
jitsi-2.11.5604-x64.exe           2018-04-14 03:22   63M
jitsi-2.11-nightly.latest-x64.exe 2018-04-14 03:22   63M  
```
* Even Nightly Builds, the Mac worked correctly.

## Steps to reproduce
* Jitsi (Nightly Builds) can reproduce 100%.
* This problem was confirmed by the following sample code that libjitsi is causing, jitsi is not a problem.
* This problem does not reproduce when commit is reverted until about February 2017, and sample code also works correctly.
* ![image](https://user-images.githubusercontent.com/41299235/42771619-c0d4f09e-8962-11e8-9e8c-40e17f9bb643.png)

```
LibJitsi.start();
List<MediaDevice> devices = ((MediaServiceImpl) LibJitsi.getMediaService()).getDevices(MediaType.VIDEO,
        MediaUseCase.CALL);
MediaDevice device = devices.get(0);

JFrame f = new JFrame();
Component comp = (Component) LibJitsi.getMediaService().getVideoPreviewComponent(device, 640, 480);
f.getContentPane().add(comp, BorderLayout.CENTER);
f.pack();
f.setVisible(true);
```

# Environment details
* Jitsi 2.11.5604 Nightly Builds
* libjitsi is the master branch (From around 2017-02) 
  * From what time this phenomenon occurs, details unknown
* Windows10 x64
  * We can not check the operation of Windows 7
* Java version 1.8.0 and above

spf-tanabe:
Hi @ibauersachs  @gpolitis @bgrozev @bbaldino 
This problem may be caused by the following pull request.
https://github.com/jitsi/libjitsi/pull/403

Because I confirmed that this commit works correctly.
```
$ git log
[...]

PR #503, opened 2019-12-16
FFmpeg: AVFilter: HFlip: Allow conversion between pixel formats.

traud:
This avoids the FFmpeg internal error message: The filters 'Parsed_buffer_0' and 'Parsed_hflip_1' do not have a common format and automatic conversion is disabled.

This happens, because the source (buffer) might use a pixel format which the filter (horizontal flip; hflip) is not able to use directly. Therefore, a conversion might be need which is enabled with the keyword scale. Because the sink (buffersink) shall have the same pixel format as the source, the same is required after the filter hflip.

closes jitsi/libjitsi#431
fixes jitsi/jitsi#540


traud:
By the way, `resources/install/debian/patches/remove-hflip` is obsolte now. When you remove that file, please, do not forget to change the file `series` as well.

traud:
Is anything blocking this?

PR #504, opened 2019-12-16
JAWT Rendering: Find shared library in Linux.

traud:
closes jitsi/jitsi#400


bgrozev:
Jenkins add to whitelist

PR #505, opened 2019-12-16
MediaStream: Re-enable incoming video.

traud:
caused by commit f60f781, undoing just the culprit

fixes jitsi/jitsi#627

### DIFF
--- a/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
+++ b/src/org/jitsi/impl/neomedia/MediaStreamImpl.java
@@ -611,7 +611,7 @@ public class MediaStreamImpl
             {
                 MediaDeviceSession deviceSession = getDeviceSession();
 
-                if (deviceSession == null || rtpTranslator != null)
+                if (deviceSession == null)
                 {
                     // Since there is no output MediaDevice to render the
                     // receiveStream on, the JitterBuffer of the receiveStream


traud:
This re-enables incoming video in Jitsi Desktop for me. And this is the culprit for sure (digged through zillions of commits). Nevertheless, I am not sure if this is the correct solution. Either this code part slipped through, is incomplete, or the logic should have been the opposite (not `!=` but `==`).

src/org/jitsi/impl/neomedia/MediaStreamImpl.java
```diff
@@ -611,7 +611,7 @@ private boolean addReceiveStream(ReceiveStream receiveStream)
             {
                 MediaDeviceSession deviceSession = getDeviceSession();
 
-                if (deviceSession == null || rtpTranslator != null)
+                if (deviceSession == null)
                 {
```
damencho: Are you sure this fixes your problem? The rtpTranslator is used only in jvb and when configured in jigasi, otherwise it is always null for Jitsi Desktop and should not matter.
This was added because of jigasi and this change my break it, I think ...

traud: Yes, definitely the cause, double and triple checked, just that condition. With that additional check, the Java Media Framework (JMF) is not creating any RTPStream for incoming video.

damencho: Hum, maybe you are right and this is not needed. As jigasi is in the same state with deviceSession not null...

damencho: Actually I'm puzzled that this change breaks anything.  The whole change is about translator and in jitsi there is no translator, so even the check above is always checking just deviceSession variable.

traud: I have no idea about that part of the source code. If you need anything, like the parameter states, to de-puzzle that, please, do not hesitate to ask. I placed a `new Throwable().printStackTrace();` at that line and got this:
```
INFO: [117] net.sf.fmj.media.Log.info() Resetting queue, last seq added: 9223372036854775806, current seq: 63488
org.jitsi.impl.neomedia.MediaStreamImpl.addReceiveStream(MediaStreamImpl.java:613)
org.jitsi.impl.neomedia.MediaStreamImpl.update(MediaStreamImpl.java:3453)
org.jitsi.impl.neomedia.rtp.translator.RTPTranslatorImpl.update(RTPTranslatorImpl.java:1099)
net.sf.fmj.media.rtp.RTPEventHandler.processEvent(RTPEventHandler.java:92)
net.sf.fmj.media.rtp.RTPEventHandler.dispatchEvents(RTPEventHandler.java:50)
net.sf.fmj.media.rtp.RTPEventHandler.run(RTPEventHandler.java:114)
INFO: [160] net.sf.fmj.media.Log.info() Starting RTPSourceStream.
INFO: [163] org.jitsi.impl.neomedia.codec.video.vp8.DePacketizer.log() Opened VP8 depacketizer
```
I guess that nobody noticed that, yet, because that more obvious HFlip issue let anyone think that video does not work in general.

damencho:
This will be fixed with https://github.com/jitsi/libjitsi/pull/507

PR #506, opened 2019-12-16
Removes h263 codec code as it is not used.

damencho:


src/org/jitsi/impl/neomedia/codec/FMJPlugInConfiguration.java
```diff
@@ -177,14 +173,6 @@ public static void registerCustomCodecs(boolean enableFfmpeg)
                 "com.sun.media.codec.video.colorspace.RGBScaler",
                 PlugInManager.CODEC);
 
-        // Remove JMF's H263 codec.
```
bgrozev: If JMF has a 263 impl we probably still want to disable it right?

damencho: We don't use JMF but fmj and I was seraching for those classes and could not find those, that's why I deleted this

bgrozev: D'uh of course...sorry for the noise

pom.xml
```diff
@@ -68,7 +68,7 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-lgpl-dependencies</artifactId>
       <optional>true</optional>
-      <version>1.1-20190327.160813-5</version>
+      <version>1.1-0-gc27d819</version>
```
bgrozev: I wonder if this is considered an older version. Should we bump to 1.2?

damencho: https://github.com/jitsi/jitsi-lgpl-dependencies/pull/14

PR #513, opened 2020-01-07
Update VP8 according to RFC 7741

lem21h:
Changes for bit:
L: TL0PICIDX present.  When set to 1, the TL0PICIDX MUST be present
      and specified as below, and the T bit MUST be set to 1.
      Otherwise, TL0PICIDX MUST NOT be present.

Add Y bit and KEYIDX value to be obtain

### DIFF
--- a/src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
+++ b/src/org/jitsi/impl/neomedia/codec/video/vp8/DePacketizer.java
@@ -418,7 +418,7 @@ public class DePacketizer
 
     /**
      * A class that represents the VP8 Payload Descriptor structure defined
-     * in {@link "http://tools.ietf.org/html/draft-ietf-payload-vp8-10"}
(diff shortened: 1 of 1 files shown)


bgrozev:
Thanks for the contribution!

Are you actively using this code? We've now moved to jitsi#jitsi-media-transform, where it's been slightly modified already: https://github.com/jitsi/jitsi-media-transform/blob/master/src/main/java/org/jitsi_modified/impl/neomedia/codec/video/vp8/DePacketizer.java

If you use the de-packetizer alone, I would suggest to migrate to jitsi-media-transform (which apart from including kotlin should be a much lighter dependency), because it is where our efforts are currently concentrated.

lem21h:
Hi, @bgrozev thank you for the update. I've been using libjitsi for a while.
I had a look into that jitsi-media-transform and the changes that I've made here are also missing there.
If you don't mind, I will make PR there as well :)

PR #515, opened 2020-01-23
AMR-WB: Enable VAD/DTX and use GSM compatible bitrate as default.

traud:
AMR-WB is the bandwidth-efficient sibling of the HD audio codec G.722. Therefore, really leverage it features and enable voice-activation detection (VAD) so the RTP packets are transmitted discontinuously (DTX). Although this works with all versions of the encoder, at least version 0.1.5 of the OpenCORE-AMR decoder is recommended (comes in Debian 11 and Ubuntu 20.04 LTS), because previous versions had a bug in the Silence Descriptor (SID) decoding which resulted in white noise. Furthermore, to reduce the bandwidth even more, do not use the highest (AMR-WB Mode 8) but the GSM compatible bitrate as default (AMR-WB Mode 2). That still gives HD audio but works not only on UMTS (3G) but even on GSM links (2G).

### DIFF
--- a/src/org/jitsi/impl/neomedia/codec/audio/amrwb/JNIEncoder.java
+++ b/src/org/jitsi/impl/neomedia/codec/audio/amrwb/JNIEncoder.java
@@ -101,7 +101,7 @@ public class JNIEncoder
     /**
      * The bit rate to be produced by this <tt>JNIEncoder</tt>.
      */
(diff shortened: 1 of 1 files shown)


traud:
The next step would be making this default configurable from Jitsi Desktop → Menu → Options → Advanced → AMR.

PR #529, opened 2020-05-20
fix: Workaround to always stop AudioSilenceStream and not hang.

damencho:


src/org/jitsi/impl/neomedia/jmfext/media/protocol/audiosilence/DataSource.java
```diff
@@ -386,11 +386,27 @@ public synchronized void stop()
...
+                    break;
+                }
+
                 try
                 {
-                    wait();
+                    long started = System.currentTimeMillis();
+                    wait(WAIT_TIMEOUT);
+                    waited
+                        = System.currentTimeMillis() - started >= WAIT_TIMEOUT;
```
bgrozev, line 409: This requires a single `wait` to last at least 100ms. I think you can just move the `started` initialization above the loop.

ibauersachs: @damencho 
`currentTimeMillis` is not guaranteed to be monotonic. If you run this and the time is adjusted by the OS/VM host, you might get a negative delta, possibly in the range of hours. Better use [System.nanoTime()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--) instead, and be aware of the possible numeric overflow mentioned in the Javadoc.

damencho: @ibauersachs Something like that: https://github.com/jitsi/libjitsi/pull/530?

PR #542, opened 2021-02-22
fix: Uses ssrc from the RTPManager when translator is used.

damencho:
In case of jigasi we are announcing the ssrc that will be sent early in the jingle and we want to return the ssrc from the translator/rtpManager and that to be the only value (not the random from the AudioMediaStreamImpl or -1 from RTPManagerImpl member localSSRC).

### DIFF
--- a/src/org/jitsi/impl/neomedia/rtp/translator/RTPTranslatorImpl.java
+++ b/src/org/jitsi/impl/neomedia/rtp/translator/RTPTranslatorImpl.java
@@ -719,7 +719,17 @@ public class RTPTranslatorImpl
      */
     public long getLocalSSRC(StreamRTPManager streamRTPManager)
     {
-         if (streamRTPManager == null)
+        // XXX(damencho) in jigasi we are announcing the ssrc and use it later
+        // to rewrite whatever comes from sip to send it to the bridge with correct ssrc.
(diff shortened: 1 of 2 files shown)


src/org/jitsi/impl/neomedia/rtp/translator/RTPTranslatorImpl.java
```diff
@@ -719,9 +719,9 @@ public GlobalTransmissionStats getGlobalTransmissionStats(
      */
     public long getLocalSSRC(StreamRTPManager streamRTPManager)
     {
-        // if (streamRTPManager == null)
-        //    return localSSRC;
-        // return ((RTPSessionMgr) manager).getLocalSSRC();
+         if (streamRTPManager == null)
```
bbaldino, line 722: I'd add a longer comment up here which explains why the assumption below was causing issues and what this is addressing (getting different SSRCs at different points in time).  I don't think I follow the comment you added below.

bbaldino: Also I think we need to adjust the comment on `localSSRC` at the top, as it documents `localSSRC` as overriding the SSRC rom the RTP manager, but now it's the opposite.

bbaldino (approved):

PR #544, opened 2021-04-22
fix: Initialize the zrtp timeout thred only when needed.

damencho:
In jigasi we saw some TimeoutProvider threads leaking ... where zrtp is not used. Seems there is a scenario in which the ZRTPTransformEngine is initialized but never stopped/cleaned. And zrtp is not used there.
Rarely happening, but still happening 20 times out of 2k.

### DIFF
--- a/src/org/jitsi/impl/neomedia/transform/zrtp/ZRTPTransformEngine.java
+++ b/src/org/jitsi/impl/neomedia/transform/zrtp/ZRTPTransformEngine.java
@@ -577,6 +577,14 @@ public class ZRTPTransformEngine
         zrtpEngine
             = new ZRtp(zf.getZid(), this, clientIdString, config, mitmMode);
 
+        if (timeoutProvider == null)
+        {
+            timeoutProvider = new TimeoutProvider("ZRTP");
+            // XXX Daemon only if timeoutProvider is a global singleton.
+            // timeoutProvider.setDaemon(true);
+            timeoutProvider.start();
+        }
+
         enableZrtp = autoEnable;
         return true;
     }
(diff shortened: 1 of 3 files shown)


src/org/jitsi/impl/neomedia/transform/zrtp/ZRTPTransformEngine.java
```diff
@@ -577,14 +577,6 @@ public synchronized boolean initialize(
         zrtpEngine
             = new ZRtp(zf.getZid(), this, clientIdString, config, mitmMode);
 
-        if (timeoutProvider == null)
```
gpolitis, line 580: This change is in the right direction but I am worried that we may break auto-sensing. How about we just check here if ZRTP is enabled? In Jigasi this is always false.

    if (autoEnable && timeoutProvider == null)

gpolitis (changes requested):

ibauersachs:
It'd probably be even better to avoid creating the ZRTP engine and SrtpControl at all. Not sure why that's happening when ZRTP is disabled. Maybe this needs to be changed to the NullControl (which doesn't do encryption):

https://github.com/jitsi/libjitsi/blob/d1f44f8eed24e3e45270e459e26f23d3e75d0ca1/src/org/jitsi/impl/neomedia/MediaStreamImpl.java#L413-L417

I'm not sure anymore what the comment just above refers to. It's from a time when libjitsi was still part of Jitsi itself.

damencho:
Yeah, I was thinking the same, why we even create it ... :)

damencho:
@gpolitis @ibauersachs  So in my tests I always had DEFAULT_ENCRYPTION=false and with this change the NullControl is used and if I comment DEFAULT_ENCRYPTION setting, leaving default zrtp is initialized as before. 
And about the comment, I saw nothing that needs to be adjusted 🤷‍♂️

gpolitis (approved):

damencho:
Thanks @ibauersachs

PR #548, opened 2021-05-24
Adds an option to clear stats (send)

damencho:


src/org/jitsi/impl/neomedia/stats/MediaStreamStats2Impl.java
```diff
@@ -140,6 +151,24 @@ public void rtcpReceiverReportReceived(long ssrc, int fractionLost)
...
+        this.cleanSendStatsOld();
+    }
+
+    /**
+     * Clean old send stats.
+     */
+    private void cleanSendStatsOld()
+    {
+        long now = System.currentTimeMillis();
+        this.sendSsrcStatsToClean.entrySet().stream().forEach(entry ->
```
gpolitis, line 164: This probably has negligible overhead but since this may run often, shall we check if the map is empty?

src/org/jitsi/impl/neomedia/stats/MediaStreamStats2Impl.java
```diff
@@ -344,6 +373,15 @@ public void removeReceiveSsrc(long ssrc)
         receiveSsrcStats.remove(ssrc);
     }
 
+    /**
+     * Clears ssrc from receiver stats.
```
gpolitis, line 377: There's a small typo here, it's the send stats that we're clearing

src/org/jitsi/service/neomedia/RawPacket.java
```diff
@@ -1888,6 +1893,24 @@ public boolean setVersion()
...
+    public boolean isSkipStats()
+    {
+        return skipStats;
+    }
+
+    /**
+     * Changes the skipStats flag.
+     * @param skipStats the new value.
+     */
+    public void setSkipStats(boolean skipStats)
```
gpolitis, line 1909: I don't see this used anywhere. Are we calling it from Jigasi?

gpolitis (approved):

src/org/jitsi/service/neomedia/RawPacket.java
damencho: Yes.

src/org/jitsi/impl/neomedia/stats/MediaStreamStats2Impl.java
damencho: I'm adding it.
But this is not running so often, it's running on every rtcp report.

gpolitis (approved):

PR #551, opened 2021-09-06
Add initial support for ppc64le

madscientist159:
This fixes https://github.com/jitsi/jigasi/issues/380


ibauersachs:
This should go into #539.

madscientist159:
@ibauersachs Is this something you could pull in to #539 or do I need to figure out how to integrate it with the WIP patches there?

ibauersachs:
I cannot easily pull this into #539 as the entire build has changed (CMake + Maven). There are also no binaries in committed anymore, they're built as part of the GitHub Actions. I didn't bother about additional architectures at the time, so it might be a bit more complicated than necessary to add another one (or two or three).

PR #549, opened 2021-07-06
ARM64 support

jbg:
With this patch I was able to build `libjnopus.so` for ARM64 and thus get Jigasi working on an ARM64 server. I was also able to build `libjnsctp.so` but haven't tested any of the others yet.

PR #552, opened 2021-11-07
Add linux-arm64 and linux-ppc64el

ibauersachs:
Closes #551 
Closes #549


Felixkruemel:
The CI at least fails with that error:
```
docker: Error response from daemon: pull access denied for libjitsi-static, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```
This does not seem to be an hard issue, but I wonder why it works in 3rd CI.

ibauersachs:
The CI error is not directly related; I was trying to things at once (speeding up the build and adding architectures).

If you build the docker image locally to run the actual build, you'll see the vpx build error on arm. Partially at least, the image build fails too.
I have some uncomitted stuff / bash history laying around to fix the package installation (the gcc-multilib package has conflicts with an :arm64 package, there's a Ubuntu bug open somewhere).

damencho (approved):

PR #554, opened 2022-01-09
Build Debian packages

ibauersachs:


damencho:
I was just wondering about jigasi, so we will need to depend on another repo when that reaches it?

damencho (approved):

ibauersachs:
> I was just wondering about jigasi, so we will need to depend on another repo when that reaches it?

No, the version released to Maven Central still contains the statically linked native libraries. I only added new builds that produce dynamically linked libs, packaged in .debs for the respective Ubuntu/Debian series. This distinction of the series is missing in the Debian repo that is currently used. This will be an issue (again) once OpenSSL v3 lands in Debian, just like we had with OpenSSL 0.9/1.0/1.1, ffmpeg and libav.

PR #559, opened 2022-05-02
ci: use env var

saghul:


damencho:
Fixed in https://github.com/jitsi/libjitsi/pull/558

PR #565, opened 2023-05-17
fix: Fix github workflows.

damencho:
ubuntu-18.04 was deprecated and is no longer available.

ubuntu-20.04 includes various PHP versions from ppa:ondrej/php (see https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/php.sh). That ppa also includes custom packaged version of libpcre*, which break standard Ubuntu packages, namely openjdk-11-jdk. Get rid of them via ppa-purge.

The runner images for ubuntu-22.04 fortunately no longer has that stuff, see actions/runner-images#6331.

### DIFF
--- a/.github/workflows/maven-and-native.yml
+++ b/.github/workflows/maven-and-native.yml
@@ -133,11 +133,14 @@ jobs:
         env:
           DEBIAN_FRONTEND: noninteractive
         run: |
-          ls -al /etc/apt/sources.list.d/
-          cat /etc/apt/sources.list
-          sudo rm -f /etc/apt/sources.list.d/ubuntu-toolchain-r-ubuntu-test-focal.list
-          sudo apt-get remove --purge g++-9 g++-10
-          sudo apt-get autoremove --purge
+          # Remove broken packages from pre-installed php in Actions image
+          # No longer needed in ubuntu-22.04
+          sudo apt-get remove -y libpcre2-dev libicu-dev icu-devtools
+          sudo apt-add-repository ppa:ondrej/php -y
+          sudo apt-get install -y aptitude
+          echo "Get::allow-downgrades \"true\";" | sudo tee /etc/apt/apt.conf.d/99-downgrades
+          echo "Get::Assume-Yes \"true\";" | sudo tee -a /etc/apt/apt.conf.d/99-downgrades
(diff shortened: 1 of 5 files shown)


ibauersachs:
Use a Docker image locally to Debug this or you'll get insane.

Also, don't copy the ports sources if you're not building for x86/x86-64.

ibauersachs:
Oh, and check if the `proposed` is still active. That has had broken dependencies for a while even in 22.04.

damencho:
Well it behaves differently in the Docker and in the github workflows ... different problems, different dependencies missing ...

damencho:
One of the problem I see on the docker is that it tries to install at the same time cpp and cpp:i386 and the second one removes the first one ...

damencho:
And seems there is more to that ... while debugging it I see "Broken libtool:arm64" ...

damencho:
And apparently I need ` --platform linux/amd64` when running on m1. I changed only `ARG ARCH=x86` in the Dockerfile (and adding `ENV DEBIAN_FRONTEND=noninteractive` because of tzdata) and  everything went fine, success. 

So the problems seems specific to the github images used for the workflows.

ibauersachs:
I'll try to have a look tonight.

damencho:
So the pulse lib is a conflict with git about libpcre2-8-0:amd64 or libpcre2-8-0:i386 .. maybe if we move to git:i386 ... :) I may try it out to see will that change anything.

damencho:
Well the same conflict comes from libselinux1 ... not sure how to solve these ...

damencho:
Can we actually drop i386 builds at all ?

damencho:
I don't see another option for now other than dropping the static building which means jitsi-desktop and jigasi will not run on x86, arm64 and ppc64el anyway we build jigasi only for amd64 ... maybe not a big deal for the moment. And fixing the windows and bullseye failures ... WDYT @ibauersachs

ibauersachs:
Not giving up quite yet, and this is definitely an issue with the GitHub actions runners. The other option is to build with Docker on the Actions.

damencho:
O yeah, you are right I forgot I ran it and install everything fine on the Docker image once I fixed the m1 shenanigan ... 
🤞

damencho:
🕺 Thank you

PR #566, opened 2023-05-21
Fix set output

ibauersachs:


damencho (approved):

damencho:
Hey @ibauersachs, do you know why it does not appear in maven central?

ibauersachs:
Publishing fails, it seems Maven Central has issues: https://status.maven.org/
It says 'resolved', but it isn't. The admin ui at https://oss.sonatype.org/ doesn't even load properly.

damencho:
👍

PR #573, opened 2024-01-24
Bump bouncycastle to version 1.77.

JonathanLennox:


JonathanLennox:
@ibauersachs The failing CI tests seem clearly unrelated to this PR - did something change in the environment?  It looks like it's having some trouble with a GPG key.  Can you take a look?

ibauersachs:
I need to check which key that was or if it might have expired.
It should be a secret at org-level, not sure why that would have changed. The audit log might give a clue.

ibauersachs:
Ah. Since this PR is from your private fork, it cannot access our org secrets. Create the PR from a branch within jitsi/libjitsi and it should be fine.

JonathanLennox:
Closed in favor of https://github.com/jitsi/libjitsi/pull/574

PR #589, opened 2025-10-31
fix builds & update dependencies

damencho:


resources/deb-build.sh
```diff
@@ -13,29 +23,43 @@ mkdir -p "${BUILD_DIR}"
...
 if [[ "${ARCH}" != "amd64" ]]; then
-  mk-sbuild "${DIST}" --target "${ARCH}" --skip-security --type=file --debootstrap-include=ca-certificates || sbuild-update -udc "${DIST}"-amd64-"${ARCH}"
+  # For non-amd64, set mirror for ports if needed
+  if [[ "${ARCH}" == "arm64" || "${ARCH}" == "ppc64el" ]]; then
+    if ubuntu-distro-info --all | grep -Fqxi "${DIST}"; then
+      export DEBOOTSTRAP_MIRROR=http://ports.ubuntu.com/ubuntu-ports
+    fi
+  fi
+
+  # Create native chroot for the target architecture (using QEMU for emulation)
```
ibauersachs, line 34: Was this the LLMs suggestion? I'm surprised this built, because when I wrote this originally the qemu builds were broken while installing Java and ca-certificates. It was also painfully slow to do anything at all, let alone produce a successful build...

damencho: Yeah, llm and this was the only successful build I managed to produce, was about to test it tomorrow.

damencho: I tested few scenarios on a linux machine and seems everything works as expected. I will proceed with updating jigasi and testing it on meet.jit.si

PR #592, opened 2026-05-08
Skip deb package signing on PR builds

damencho:
Secrets (GPG_KEY, GPG_PW) aren't available on PRs from forks, causing the GPG import step to fail with "no valid OpenPGP data found". Guard the import on master only and skip debsign when no GPG ID is provided so PRs still validate the deb build.

### DIFF
--- a/.github/workflows/maven-and-native.yml
+++ b/.github/workflows/maven-and-native.yml
@@ -201,7 +201,7 @@ jobs:
       - name: Install tools
         run: |
           # No need for updated man pages on GitHub
-          sudo rm /var/lib/man-db/auto-update
+          sudo rm -f /var/lib/man-db/auto-update
           resources/deb-prepare.sh
 
       - name: Import GPG key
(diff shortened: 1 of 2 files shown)


JonathanLennox (approved):

.github/workflows/maven-and-native.yml
```diff
@@ -221,7 +222,7 @@ jobs:
             "${{ needs.version.outputs.version }}" \
             "${{ matrix.dist.dist }}" \
             "${{ matrix.arch }}" \
-            "dev+maven@jitsi.org"
+            "${{ github.ref == 'refs/heads/master' && 'dev+maven@jitsi.org' || '' }}"
```
ibauersachs, line 225: The && here is weird. The e-mail was a parameter to the script.

ibauersachs (changes requested):

.github/workflows/maven-and-native.yml
damencho: in case of master it will pass it, otherwise it will be empty string which will skip the debsign.

JonathanLennox: The github action syntax is stupid but that's how you do a conditional string replace.  See https://stackoverflow.com/questions/69570225/conditionally-append-a-string-in-a-github-actions-workflow

